PR #32, opened 2025-05-08
Upgrade MEDS transforms to v0.4.0

mmcdermott:
This doesn't handle the MEDS upgrade yet; that will come separately to keep things simple.

This is not yet working. It needs additional work to get the extraction pipeline working and to get the tests to pass, let alone to use the proper stage example format.


mmcdermott:
@rvandewater this PR now passes tests and uses the newer MEDS-transforms; however, I dropped the two larger integration tests and now this system lacks a clear way to run the full pipeline. The full pipeline configuration still exists, and could be corrected with just a documentation update, but I wanted to know how urgently you need this upgrade pushed vs whether or not we should integrate other improvements. Thoughts?

Issue #81, opened 2026-04-22
Integrate download functionality as a pseudo-stage in the pipeline


mmcdermott:
<html><head></head><body><h1>Shared download module: design proposal for <code>meds_extract.download</code></h1>
<p><strong>Intended repo:</strong> <code>MEDS_extract</code> (primary — this is the proposed home). Cross-referenced from <code>ETL_MEDS_Template</code> and each ETL's issue tracker. Depends on Phase 0 quick-wins shipping first so this can proceed without user pressure.</p>
<p><strong>One-line summary:</strong> move download into <code>MEDS_extract</code> as a first-class <code>Source</code> abstraction with pluggable backends (PhysioNet, HTTP, fsspec, Dataverse). Delete <code>download.py</code> from every ETL.</p>
<hr>
<h2>Why this exists</h2>
<p>Six of seven public ETLs have a near-identical <code>download.py</code>, copy-inherited from <code>ETL_MEDS_Template</code>. A single bug (issue #42: no timeouts) therefore exists in N places and has to be patched in N places. The template is a GitHub "template repo" — a one-shot clone — so there is no mechanism to push a fix to derivatives. The quick-wins patch (doc 01) demonstrates this pain concretely: the same diff has to land in MIMIC, eICU, HIRID, and the template as four separate PRs.</p>
<p>Simultaneously, the current code cannot handle the full space of sources the ecosystem actually has:</p>
<ul>
<li>PhysioNet with Basic auth (MIMIC-IV, eICU, HiRID, NWICU, SICdb, INSPIRE)</li>
<li>PhysioNet open demos (no auth)</li>
<li>Public GitHub raw URLs (MIMIC's <code>concept_map</code> shared metadata)</li>
<li>DANS DataVerse with API key (AUMCdb)</li>
<li>Cloud buckets (future; a dataset hosted on S3/GCS would be entirely unserved today)</li>
<li>Local filesystem (for re-running with already-downloaded data, which currently requires <code>do_download=False</code> and manual placement)</li>
</ul>
<p>The <code>requests</code> + BeautifulSoup HTML-crawl pattern doesn't generalize to any of these cleanly. Each new source shape currently costs a new <code>download.py</code> fork.</p>
<h2>Where it lives — and why in <code>MEDS_extract</code></h2>
<h3>The candidates I considered</h3>

Option | Pros | Cons
-- | -- | --
[...]

PR #82, opened 2026-04-22
Add meds_extract.download: Source / Fetcher abstraction (refs #81)

mmcdermott:
Refs #81.

## Summary

Initial cut at the download pseudo-stage described in #81. Delivers the core
`meds_extract.download` abstraction so the six+ public ETLs (MIMIC-IV, eICU, HIRID,
NWICU, SICdb, INSPIRE, AUMCdb) can eventually delete their near-identical forks of
`download.py` + `dataset.yaml` and declare a `sources:` block in their MESSY configs instead.

## What's in

- **`Source` protocol** + `RemoteFile` / `FetchResult` / `FetchReport` dataclasses
- **`Fetcher`** — bounded-concurrency orchestrator (ThreadPoolExecutor, configurable `max_concurrency`, optional `continue_on_error`)
- **Three backends**:
  - `HTTPSource` — explicit URL list with optional per-URL SHA-256 + `rel_path` override (MIMIC's `common:` block of GitHub-hosted concept-map CSVs)
  - `FsspecSource` — any UPath-addressable root (`file://`, `s3://`, `gs://`, …) for re-runs against pre-downloaded mirrors
  - `PhysioNetSource` — `SHA256SUMS.txt`-manifest-driven (no BeautifulSoup crawl!), HTTP Basic auth for restricted datasets
- **`source_from_config` / `sources_from_spec`** — read a MESSY `sources:` block
- **`meds-extract-download`** console script — Hydra-driven, override conventions match the rest of the pipeline

Core primitives:

- **`_resumable_download`** — `.part`-staging + Range-based resume + SHA-256 verify + atomic rename. Handles server-ignores-Range (200 on Range request → restart from byte 0). Stale local files (wrong SHA-256) get re-fetched.
- **`_make_httpx_client`** — wraps `httpx.Client` with a `tenacity` retry policy (connect/read/write timeouts, 5xx, `RemoteProtocolError`; 4xx is not retried — wrong URL / bad auth fails immediately).

## Dependencies

- `httpx>=0.27` + `tenacity>=8` gated behind a `download` extra (`pip install 'MEDS-extract[download]'`). No runtime impact on transform-only users.

## Tests

**98 passing** (28 new for the download layer):

- **13 integration tests** via `httpx.MockTransport` — Range resume, server-ignores-Range restart, checksum mismatch + `.part` cleanup, stale-dest re-fetch, `HTTPSource` / `PhysioNetSource` / `Fetcher` end-to-end, `continue_on_error` fault tolerance.
- **15 doctests** on every helper: `_sha256`, `_make_httpx_client`, dispatch, `SHA256SUMS.txt` parsing (including comments + blank lines + spaced paths), URL normalization, `Fetcher._already_complete` across size/sha256 match+mismatch cases.

**Coverage**: 100% on `_http.py`, `source.py`, `fetcher.py`, `dispatch.py`, `backends/http.py`, `backends/physionet.py`; 93% on `backends/fsspec.py` (one error-path in a `finally`); `cli.py` at 39% (Hydra entry-point plumbing — untestable without a subprocess harness, follow-up).

## Deliberately out of scope (per #81)

- `DataverseSource` (AUMCdb's DANS DataVerse API) — next PR
- Manifest of downloaded files on disk for fast re-run skip (v2)
- Pydantic validation for `sources:` entries
- Actually migrating the ETL-side `download.py` files — those are per-repo PRs using this library
- The `dataset.yaml` → MESSY unification inside the ETL templates — cross-repo change


src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,189 @@
...
+import logging
+from typing import TYPE_CHECKING
+
+try:
+    import httpx
+except ImportError as e:  # pragma: no cover
+    raise ImportError(
+        "MEDS_extract.download requires the 'download' extra. "
+        "Install with: pip install 'MEDS-extract[download]'"
+    ) from e
```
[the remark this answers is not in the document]
mmcdermott: Fixed in afb6ffa. Both `ImportError` messages + the `__init__.py` docstring now say `pip install 'MEDS_extract[download]'`, matching `name = "MEDS_extract"` in `pyproject.toml`. (Pip normalizes both `MEDS_extract` and `MEDS-extract` per PEP 503, but the hint should match the source of truth.)

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,189 @@
...
+        retry,
+        retry_if_exception_type,
+        stop_after_attempt,
+        wait_exponential,
+    )
+except ImportError as e:  # pragma: no cover
+    raise ImportError(
+        "MEDS_extract.download requires the 'download' extra. "
+        "Install with: pip install 'MEDS-extract[download]'"
+    ) from e
```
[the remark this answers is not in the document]
mmcdermott: Fixed in the same commit (afb6ffa) — same one-character edit applies to both `ImportError` messages.

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,189 @@
...
+        follow_redirects=True,
+    )
+
+    retry_decorator = retry(
+        stop=stop_after_attempt(max_retries),
+        wait=wait_exponential(multiplier=1, min=1, max=30),
+        retry=retry_if_exception_type(_RETRY_EXC),
+        reraise=True,
+    )
+    client.get = retry_decorator(client.get)  # type: ignore[method-assign]
```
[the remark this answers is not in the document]
mmcdermott: Real issue, fixed in afb6ffa by wrapping a helper that calls `raise_for_status()` inside the retry boundary for **5xx only**:

```python
def _get_with_5xx_retry(*args, **kwargs):
    response = original_get(*args, **kwargs)
    if 500 <= response.status_code < 600:
        response.raise_for_status()
    return response

client.get = retry_decorator(_get_with_5xx_retry)
```

Kept 4xx out of the retry path (bad URL / bad auth shouldn't spam retries). Added a doctest that passes a `MockTransport` returning 503, 503, 200 and verifies it takes exactly 3 attempts to reach the 200.

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,189 @@
...
+    # Can't wrap `stream` the same way — it returns a context manager, not a response. See
+    # `_resumable_download` for retry handling on streaming downloads.
+    return client
```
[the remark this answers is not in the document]
mmcdermott: Fixed in afb6ffa by adjusting the docstring to match actual behavior (second half of your suggestion). Implementing streaming retry with mid-stream resume is non-trivial and risks masking real errors; the current contract is clearer:

- Non-streaming `client.get` retries (including 5xx) — this covers the manifest fetch in `PhysioNetSource`, which is what typically faces 5xx blips.
- Streaming read inside `_resumable_download` surfaces mid-stream errors to the caller; callers who want retry-across-the-whole-file can wrap their `Fetcher.fetch_all` call in tenacity and the `.part` file will be picked up on the re-invocation via `Range: bytes=N-`.

Module docstring now explains this explicitly so there's no hidden promise.

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Examples:
+            >>> import tempfile, hashlib
+            >>> from pathlib import Path
+            >>> with tempfile.TemporaryDirectory() as d:
+            ...     d = Path(d)
+            ...     fp = d / "x.txt"
+            ...     fp.write_bytes(b"abc")
+            ...     Fetcher._already_complete(fp, RemoteFile("x.txt"))  # no manifest info
+            3
```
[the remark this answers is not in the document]
mmcdermott: Fixed in afb6ffa. Applied your suggestion — `_ = fp.write_bytes(b"abc")` suppresses the `3` return value, and the expected output is now just `True`.

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,196 @@
...
+    """Drives a :class:`Source` through a bounded-concurrency download plan.
+
+    Attributes:
+        dest_dir: Root directory that :class:`RemoteFile` ``rel_path`` values land under.
+        max_concurrency: Maximum number of parallel transport streams. 4 is polite against
+            rate-limiting servers (PhysioNet); 8 is typically safe; 16+ risks a ban.
+        continue_on_error: If ``True``, per-file transport exceptions are captured as
+            :class:`~MEDS_extract.download.source.FetchResult` with ``status="failed"`` and
+            the run proceeds. If ``False`` (default), the first failure cancels remaining
+            in-flight transfers and re-raises.
```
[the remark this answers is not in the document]
mmcdermott: Fixed in afb6ffa by updating the docstring to match actual behavior (your second suggestion). `ThreadPoolExecutor` doesn't cancel already-submitted futures, and adding that complexity isn't worth it here — the docstring now says:

> If `False` (default), the first failure is re-raised; already-submitted concurrent transfers are **not** explicitly canceled (they complete in the background and their results are discarded).

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,196 @@
...
+        logger.info(
+            f"Fetch complete: {report.n_downloaded} downloaded, {report.n_skipped} skipped, "
+            f"{report.n_failed} failed."
+        )
+        return report
+
+    def _fetch_one(self, source: Source, remote: RemoteFile) -> FetchResult:
+        dest = self.dest_dir / remote.rel_path
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        if self._already_complete(dest, remote):
```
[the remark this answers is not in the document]
mmcdermott: Real security issue, fixed in afb6ffa. `Fetcher._fetch_one` now routes through a new `_resolve_dest(rel_path)` method that:

1. Rejects absolute paths outright (`ValueError: rel_path must be relative, got absolute`).
2. Resolves the full path under `dest_dir` and checks `resolved.relative_to(dest_root)` — any `..` escape trips the `ValueError`.

Added two integration tests (`test_fetcher_rejects_path_traversal` with `../../etc/passwd` and `test_fetcher_rejects_absolute_path` with `/etc/passwd`) plus doctests on `_resolve_dest` itself covering the happy path, absolute-path rejection, and `..` escape rejection.

src/MEDS_extract/download/backends/physionet.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(
+        self,
+        base_url: str,
+        username: str | None = None,
+        password: str | None = None,
+        client: httpx.Client | None = None,
+    ):
+        self._base_url = base_url if base_url.endswith("/") else base_url + "/"
+        auth = (username, password) if username else None
```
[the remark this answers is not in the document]
mmcdermott: Applied your suggestion (slightly extended) in afb6ffa. `PhysioNetSource.__init__` now raises early:

```python
if (username is None) != (password is None):
    raise ValueError(
        f"{type(self).__name__}: username and password must be supplied together "
        f"(got username={username!r}, password={'<set>' if password else None!r}). "
        f"Omit both for open-access datasets (e.g. MIMIC-IV demo)."
    )
```

The `'<set>'` placeholder for the password avoids leaking the actual password into the error message. Added `test_physionet_source_rejects_half_credentials` covering both directions (username-only and password-only).

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Attributes:
+        dest_dir: Root directory that :class:`RemoteFile` ``rel_path`` values land under.
+        max_concurrency: Maximum number of parallel transport streams. 4 is polite against
+            rate-limiting servers (PhysioNet); 8 is typically safe; 16+ risks a ban.
+        continue_on_error: If ``True``, per-file transport exceptions are captured as
+            :class:`~MEDS_extract.download.source.FetchResult` with ``status="failed"`` and
+            the run proceeds. If ``False`` (default), the first failure is re-raised;
+            already-submitted concurrent transfers are **not** explicitly canceled (they
+            complete in the background and their results are discarded).
```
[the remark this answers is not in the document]
mmcdermott: Right — `ThreadPoolExecutor.__exit__` calls `shutdown(wait=True)`, so `fetch_all()` does block on submitted workers even after re-raising. Applied your suggestion in 7bd68c4; docstring now reads:

> already-submitted concurrent transfers are **not** explicitly canceled, so `fetch_all()` may still block until submitted worker tasks finish (the `ThreadPoolExecutor` context manager calls `shutdown(wait=True)` on exit) and their results are then discarded.

Considered implementing actual cancellation (e.g. `shutdown(cancel_futures=True)` on Python 3.9+) but it'd only cancel pending — already-running transfers can't be interrupted mid-stream — so the wait would still happen for whatever fraction of the pool was already in-flight. Not worth the API complexity given the typical shape (small max_concurrency, mostly large files).

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Args:
+        auth: Optional ``(username, password)`` for Basic auth — e.g. PhysioNet credentials.
+        timeout: ``(connect_timeout, read_timeout)`` in seconds.
+        max_retries: How many times to retry on transient failures before giving up.
+
+    The returned client has the same public interface as a plain :class:`httpx.Client` but
+    its ``get`` and ``stream`` methods transparently retry on connection / read-timeout /
+    5xx errors with exponential backoff. 4xx errors (wrong URL, bad auth) surface
+    immediately.
```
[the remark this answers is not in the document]
mmcdermott: Applied your suggestion in 7bd68c4. Docstring now says:

> Build an `httpx.Client` with a tenacity-wrapped `get`. […] `get` method transparently retries on connection / read-timeout / 5xx errors with exponential backoff. 4xx errors (wrong URL, bad auth) surface immediately. Streaming downloads (`client.stream`) are **not** wrapped here — they go through `_resumable_download`, which surfaces mid-stream errors to the caller and relies on the `.part` file + `Range: bytes=N-` for retry-across-the-whole-file via re-invocation.

src/MEDS_extract/download/cli.py
```diff
@@ -0,0 +1,92 @@
...
+    - ``key=dataset`` (default) / ``key=demo`` — which ``sources:`` bucket to pull.
+      ``common`` is always appended.
+    - ``concurrency=4`` (default) — max parallel transport streams per fetcher.
+    - ``continue_on_error=False`` (default) — if True, per-file failures don't sink the run.
+
+    Returns:
+        ``0`` on full success (all files downloaded / skipped, none failed), ``1`` otherwise.
+    """
+    spec_fp = Path(cfg.spec).expanduser().resolve()
+    raw_input_dir = Path(cfg.raw_input_dir).expanduser().resolve()
```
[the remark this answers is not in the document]
mmcdermott: Real footgun, fixed in 7bd68c4. `spec` and `raw_input_dir` now route through `hydra.utils.to_absolute_path()` before `expanduser().resolve()`, so `meds-extract-download spec=relative.yaml raw_input_dir=./out` resolves against the user's original CWD instead of Hydra's per-run output dir. Added an explanatory comment above the resolution so future readers don't accidentally regress this when refactoring.

mmcdermott:
### Note: why we don't use the `physionet` PyPI package

Considered and passed on integrating [MIT-LCP/physionet](https://github.com/MIT-LCP/physionet) (`pip install physionet`, currently v0.1.5). Posting the reasoning so a future reader doesn't re-ask.

**What it offers:** `PhysioNetClient.projects.list_published()` / `list_versions()` / `get_details()` / `download_checksums()` — a PhysioNet-API metadata client; a separate CLI validator for pre-submission dataset QA; auth via env vars.

**Why it's not a fit for this PR:**

1. **Doesn't solve our hard problem.** Our core primitive is streaming, resumable, SHA-256-verified large-file download (`_resumable_download` with `.part` staging + `Range: bytes=N-`). The package shows no equivalent — only `download_checksums`, which would replace ~2 lines of our `SHA256SUMS.txt` fetch. The actual MIMIC-IV-sized file streaming we'd still implement ourselves.
2. **Heavy deps.** Pulls in `numpy`, `openpyxl`, `pandas`, `requests`, `tqdm` (transitive for the validator + tabular utilities), ~50MB+ install. Our `download` extra is currently just `httpx` + `tenacity` — ~3MB.
3. **Wrong HTTP stack.** It uses `requests`; we use `httpx` for its streaming + typed API + native retry composition. Adopting would double the HTTP stack in the dep tree.
4. **Young and fluid.** First release was `0.0.0` in early 2025, latest `0.1.5`. 0.x versions signal API instability by convention.
5. **Couples config to PhysioNet's API shape.** Our `PhysioNetSource(base_url="https://physionet.org/files/mimiciv/3.1/")` works against any URL matching PhysioNet's standard layout — including demo buckets, mirrors, custom deployments. Switching to `slug="mimiciv", version="3.1"` would tie us to the live PhysioNet API's lookup shape.

**Revisit if:** the package adds a `download_project(slug, version, dest, resume=True)` primitive with streaming + `Range` + hash verification, pins to `httpx` or exposes a transport hook. Today it's a metadata-API wrapper, not a download library.

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,74 @@
...
+                extra={"upath": p},
+            )
+
+    def fetch(self, remote: RemoteFile, dest: Path) -> None:
+        upath = remote.extra["upath"]
+        part = dest.with_name(dest.name + ".part")
+        try:
+            with upath.open("rb") as src, part.open("wb") as dst:
+                shutil.copyfileobj(src, dst, length=1024 * 1024)
+            part.replace(dest)
```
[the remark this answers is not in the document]
mmcdermott: Real correctness issue, fixed in 7cd9753. `FsspecSource.fetch` now computes the SHA-256 of the staged `.part` file and raises `ChecksumError` + deletes the `.part` on mismatch, matching the HTTP/PhysioNet backends' contract.

Also refactored `_sha256` and `ChecksumError` out of `_http.py` into `source.py` (renamed to `sha256_of` for public-API clarity) so FsspecSource doesn't have to pull the httpx/tenacity dep chain through `_http.py` just to get hashing utilities.

Added `test_fsspec_source_verifies_sha256` covering both match (succeeds) and mismatch (raises, dest not created, `.part` cleaned up) cases.

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,231 @@
...
+        r.raise_for_status()
+        # Server ignored the Range header (returned 200 instead of 206) → restart.
+        if resume_from and r.status_code == 200:
+            logger.info(f"Server ignored Range for {url}; restarting from byte 0.")
+            resume_from = 0
+        mode = "ab" if resume_from else "wb"
+        with part.open(mode) as f:
+            for chunk in r.iter_bytes(chunk_size=chunk_size):
+                f.write(chunk)
+
```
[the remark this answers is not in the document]
mmcdermott: Real issue, fixed in 7cd9753. Wrapped the stream block in a `while True: ... break` loop that handles the three restart cases uniformly:

- **416 Range Not Satisfiable** (remote file shrank or changed): clear `.part`, reset `resume_from=0`, restart.
- **200 despite Range header** (server ignored Range): clear `.part`, reset `resume_from=0`, restart. (Previously kept the stale `.part` around on disk, which would have confused the next invocation.)
- **206 with mismatched Content-Range**: parse via a new `_content_range_starts_at(header, expected_start)` helper; if the server's reported start byte doesn't match our `resume_from`, clear `.part` and restart.

Added three integration tests (`test_resumable_download_mismatched_content_range_restarts`, `test_resumable_download_416_restarts`, and the existing `test_resumable_download_range_ignored_restarts_from_scratch` updated to verify `.part` is cleared) plus doctests on `_content_range_starts_at` covering malformed / missing / unsatisfied-range forms.

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,257 @@
...
+        retries):
+
+        >>> import httpx as _httpx
+        >>> attempts = []
+        >>> def flaky_then_ok(request):
+        ...     attempts.append(None)
+        ...     return _httpx.Response(503 if len(attempts) < 3 else 200, text="ok")
+        >>> client = _make_httpx_client(max_retries=5)
+        >>> client._transport = _httpx.MockTransport(flaky_then_ok)
+        >>> r = client.get("https://example.com/x")
```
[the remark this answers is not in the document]
mmcdermott: Applied your suggestion in a9fd7a8. Added an optional `transport: httpx.BaseTransport | None = None` parameter to `_make_httpx_client` that threads through to `httpx.Client(transport=...)`. The doctest now uses the public API:

```python
>>> client = _make_httpx_client(max_retries=5, transport=_httpx.MockTransport(flaky_then_ok))
```

No more reaching into `_transport` private attr. Also docs the `transport=` param as the intended test stub-point so future readers don't regress.

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        return report
+
+    def _fetch_one(self, source: Source, remote: RemoteFile) -> FetchResult:
+        dest = self._resolve_dest(remote.rel_path)
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        if self._already_complete(dest, remote):
+            logger.debug(f"Skipping {remote.rel_path}: already complete.")
+            return FetchResult(remote, dest, "skipped")
+        try:
```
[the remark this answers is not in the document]
mmcdermott: Real silent-corruption path, fixed in a9fd7a8. `Fetcher._fetch_one` now deletes `dest` before calling `source.fetch` if the file exists but failed `_already_complete` — otherwise `_resumable_download`'s `dest.exists()` early-return would leave the stale file in place and we'd still report `status="downloaded"`.

Added `test_fetcher_refetches_wrong_size_file_without_sha` as a regression guard: manifest says `size=28`, dest contains 14 bytes of wrong data, no SHA-256 provided — verifies dest is refetched to the correct content, not left stale.

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,257 @@
...
+        "Install with: pip install 'MEDS_extract[download]'"
+    ) from e
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+# Transports consider these retriable. 4xx (non-429) errors are NOT retried — those mean the
```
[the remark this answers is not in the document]
mmcdermott: Fixed in a9fd7a8 (your first suggestion — update the comment to reflect reality, since no code path currently retries 429). The comment now reads:

```
# Transports consider these retriable. 4xx errors are NOT retried (including 429) — those
# usually mean the URL is wrong or auth failed, and retrying makes things worse. If
# per-endpoint 429-with-Retry-After handling is needed later, wire it in as a separate
# retry predicate rather than expanding this list.
```

If PhysioNet or another backend starts rate-limiting us in practice, a proper
`retry_if_result(lambda r: r.status_code == 429)` predicate with `Retry-After`-aware
backoff is the right follow-up — left as a deliberate non-goal for this PR.

.github/workflows/tests.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - name: Run tests
         run: |
-          uv run pytest -v --cov=src --cov-report=xml
+          uv run --extra download pytest -v --cov=src --cov-report=xml
```
mmcdermott: If this requires extra dependencies, can you do it in two test passes, where the extras are not installed in one and are in the other? That will verify that the system still works on other paths without these things installed.

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+    """A :class:`Source` backed by an fsspec-compatible root via :class:`upath.UPath`.
+
+    Accepts any protocol UPath supports: ``file://`` and local paths (re-run against a
+    pre-downloaded copy), ``s3://`` / ``gs://`` / ``azure://`` (mirrors the user keeps on
+    cloud storage), and so on. The ``fsspec`` extras for cloud protocols (``s3fs``,
+    ``gcsfs``, …) are NOT declared as dependencies here — users install them themselves
+    following the standard fsspec pattern.
+
+    Examples:
+        >>> import tempfile
```
mmcdermott: Move this to the doctest namespace by default via a pytest fixture, to avoid cluttering doctest code. Ditto for Path.

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+        >>> from pathlib import Path
+        >>> with tempfile.TemporaryDirectory() as src_dir:
```
mmcdermott: This should use yaml_to_disk https://github.com/mmcdermott/yaml_to_disk to clean up the doctest. It should be in the namespace by default.

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+        ['labs/vitals.csv', 'patients.csv']
+
+        Fetching copies to ``dest`` atomically via a ``.part`` file:
+
+        >>> with tempfile.TemporaryDirectory() as src, tempfile.TemporaryDirectory() as dst:
+        ...     src, dst = Path(src), Path(dst)
+        ...     _ = (src / "x.txt").write_text("hello")
+        ...     source = FsspecSource(root=str(src))
+        ...     [remote] = list(source.list_files())
+        ...     dest = dst / remote.rel_path
```
mmcdermott: use pretty-print output directory to display the full contents of dst after, that way the test shows that nothing else is there (or that something else is there that is supposed to be) https://github.com/mmcdermott/pretty-print-directory

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+        upath = remote.extra["upath"]
+        part = dest.with_name(dest.name + ".part")
+        try:
+            with upath.open("rb") as src, part.open("wb") as dst:
+                shutil.copyfileobj(src, dst, length=1024 * 1024)
+            # Honor remote.sha256 the same way HTTP-backed Sources do — a mismatch is a
+            # hard error regardless of transport. Important for local-mirror re-runs:
+            # if the mirror is silently out of sync with the authoritative manifest, we
+            # want to fail loudly rather than feed corrupt data downstream.
+            if remote.sha256 is not None:
```
mmcdermott, line 86: Do you want to check if the part file exists first, and validate its sha256 in advance, then if it matches just skip?

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+                continue
+            yield RemoteFile(
+                rel_path=p.relative_to(self._root).as_posix(),
+                size=p.stat().st_size,
+                extra={"upath": p},
+            )
+
+    def fetch(self, remote: RemoteFile, dest: Path) -> None:
+        upath = remote.extra["upath"]
+        part = dest.with_name(dest.name + ".part")
```
mmcdermott, line 78: What if dest exists? Do you want a do_overwrite here? Or a general do_overwrite on the class?

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,83 @@
...
+    Examples:
+        >>> import tempfile
+        >>> from pathlib import Path
+        >>> with tempfile.TemporaryDirectory() as src_dir:
+        ...     src_dir = Path(src_dir)
+        ...     _ = (src_dir / "patients.csv").write_text("patient_id,dob\\n1,2000-01-01")
+        ...     (src_dir / "labs").mkdir()
+        ...     _ = (src_dir / "labs" / "vitals.csv").write_text("pid,hr\\n1,80")
+        ...     source = FsspecSource(root=str(src_dir))
+        ...     rel_paths = sorted(r.rel_path for r in source.list_files())
```
mmcdermott: If you add a simple str formatter on RemoteFile, then you can make this a bit more expressive without sacrificing doctest readability.

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,123 @@
...
+from ..source import RemoteFile
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    import httpx
+
+
+class HTTPSource:
+    """A :class:`Source` backed by an explicit list of HTTP URLs.
```
mmcdermott, line 49: Should Source be a base class or a protocol or something?

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,123 @@
+"""Explicit-URL-list HTTP backend."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+from urllib.parse import urlparse
+
+from .._http import _make_httpx_client, _resumable_download
```
mmcdermott: Why are these in a different file? And why not just static methods on the class (here and for the other _ methods below)?

src/MEDS_extract/download/backends/physionet.py
```diff
@@ -0,0 +1,130 @@
...
+from ..source import RemoteFile
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from pathlib import Path
+
+    import httpx
+
+
+class PhysioNetSource:
```
mmcdermott: Given this just is a limited wrapper around httpx, should it explicitly subclass it?

src/MEDS_extract/download/_http.py
```diff
@@ -0,0 +1,266 @@
...
+
+try:
+    from tenacity import (
+        retry,
+        retry_if_exception_type,
+        stop_after_attempt,
+        wait_exponential,
+    )
+except ImportError as e:  # pragma: no cover
+    raise ImportError(
```
mmcdermott: Why two separate try except blocks? And if you take my CI suggestion, you'll be able to drop the pragma no cover.

conftest.py
```diff
@@ -1,21 +1,24 @@
...
+"""
 
-import polars as pl
-import pytest
+from __future__ import annotations
 
+collect_ignore_glob = []
 
-@pytest.fixture(scope="session", autouse=True)
-def _setup_doctest_namespace(doctest_namespace: dict[str, Any]):
```
mmcdermott, line 13: Why did you remove this? WE want this.

conftest.py
```diff
@@ -1,21 +1,24 @@
...
-            "datetime": datetime,
-            "tempfile": tempfile,
-        }
+try:
+    import httpx  # noqa: F401
+    import tenacity  # noqa: F401
+except ImportError:
+    collect_ignore_glob.extend(
+        [
+            "src/MEDS_extract/download/*.py",
```
mmcdermott, line 49: What is this doing?

src/MEDS_extract/download/cli.py
```diff
@@ -0,0 +1,95 @@
...
+        continue_on_error=cfg.get("continue_on_error", False),
+    )
+    all_ok = True
+    for source in sources:
+        report = fetcher.fetch_all(source)
+        all_ok = all_ok and report.ok
+    return 0 if all_ok else 1
+
+
+def main() -> None:  # pragma: no cover — thin wrapper for the console script
```
mmcdermott: This should be tested, if it is needed, and I don't think it is -- there is a hydra_registered_dataclass utility that you can probably use instead: https://github.com/mmcdermott/MEDS_transforms/blob/03a80207e6f387461cc3cf48b5702008841c2bed/src/MEDS_transforms/configs/utils.py#L88

Moreover, I need to see examples of how you plan to use this. The design is that this information should live as part of the MESSY file, and then the main CLI entry point for a MEDS-Extract pipeline should run this first before doing the MEDS-transforms stages for the extraction. Does that still happen? Or is that offloaded to the downstream codebases? We need to clarify usage mode before proceeding here.

.github/workflows/tests.yaml
mmcdermott: Split into two jobs in df0fb8c (the single-matrix-pass approach would duplicate non-download tests, per your "no need to run the main tests again with the download extra"):

- `run_tests_ubuntu` (core): no `--extra download`. `conftest.py`'s `collect_ignore_glob` skips `src/MEDS_extract/download/**` + `tests/test_download.py` when `httpx`/`tenacity` aren't importable. Verified locally: 70 tests pass, proving the base install is complete.
- `run_tests_download`: `uv run --extra download pytest -v tests/test_download.py src/MEDS_extract/download/`. 37 tests pass locally.

Both on 3.11 and 3.12. Codecov gets separate `core` / `download` flags.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Done in 3429819. Restored the `_setup_doctest_namespace` fixture in `conftest.py` with `json`, `pl`, `datetime`, `tempfile`, and now `Path`. `yaml_disk` and `print_directory` auto-register via their own pytest plugins. The fsspec doctests no longer need any `import` lines.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Done in 3429819. First fsspec doctest now uses `yaml_disk` for tree setup:

```python
>>> spec = '''
... patients.csv: |
...   patient_id,dob
...   1,2000-01-01
... labs:
...   vitals.csv: |
...     pid,hr
...     1,80
... '''
>>> with yaml_disk(spec) as src_dir:
...     source = FsspecSource(root=str(src_dir))
...     for r in sorted(source.list_files(), key=lambda r: r.rel_path):
...         print(r)
labs/vitals.csv size=11
patients.csv size=27
```

Much easier to read than the hand-rolled `mkdir`/`write_text` version. `yaml_disk` is in the doctest namespace automatically.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Done in 3429819. Second fsspec doctest now uses `print_directory(dst)` after the fetch so the output shows the full dest tree — any unintended stray write would show up as an extra line in the ASCII tree:

```python
>>> with yaml_disk("x.txt: hello") as src, tempfile.TemporaryDirectory() as dst:
...     dst = Path(dst)
...     source = FsspecSource(root=str(src))
...     [remote] = list(source.list_files())
...     source.fetch(remote, dst / remote.rel_path)
...     (dst / "x.txt").read_text()
...     print_directory(dst)
'hello'
└── x.txt
```

The `└── x.txt` line also implicitly verifies no `.part` file leaked.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Declined per our chat discussion — going with the "Fetcher-level is the single source of skip truth" approach. Adding a `.part`-SHA pre-check inside `FsspecSource.fetch` is an optimization for the narrow case of "prior run was interrupted mid-copy, `.part` happens to be complete-and-correct." For fsspec specifically (local / cloud FS copy), the re-copy is cheap enough that this rarely pays for itself, and duplicating the skip-check in each backend risks divergence from `Fetcher._already_complete`.

If this turns out to be hot for S3-backed mirrors, happy to add it as a `Fetcher.resume_from_part` opt-in flag in a follow-up — but let's wait for evidence.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Added in a88a42c. `do_overwrite` is now threaded all the way through per our discussion:

- **CLI** (`meds-extract-download`): `do_overwrite=true` override — shows up in the Hydra config help.
- **Fetcher**: new `do_overwrite` attribute. When True, skips the `_already_complete` check, unlinks any existing `dest`, and forwards to `source.fetch(..., do_overwrite=True)`.
- **Source protocol**: `fetch(remote, dest, do_overwrite=False)` is now part of the contract.
- **Every backend's `fetch`** (`HTTPSource`, `PhysioNetSource` via inheritance, `FsspecSource`): honors the flag defensively — clears `dest` + `.part` before fetching.

Fetcher doctest covers the re-fetch-despite-match path; the end-to-end CLI test in 01dc077 exercises `do_overwrite=true` via a local modification that gets blown away on the next run.

Didn't end up needing a base class for this — adding the kwarg to three `fetch` methods is less code than a shared base would introduce, and `PhysioNetSource(HTTPSource)` already covers the natural inheritance relationship.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Added in 3429819. `RemoteFile.__str__` is now:

```python
def __str__(self) -> str:
    parts = [self.rel_path]
    if self.size is not None:
        parts.append(f"size={self.size}")
    if self.sha256 is not None:
        parts.append(f"sha256={self.sha256[:12]}...")
    return " ".join(parts)
```

Output shape:
- `x.csv`
- `x.csv size=1234`
- `patients.csv.gz size=1234 sha256=abc123def456...`

The 12-hex truncation keeps doctests readable without needing full 64-char fixtures. Used in the `FsspecSource.list_files` doctest (which now reads `labs/vitals.csv size=11` etc instead of bare rel_path strings) and landed regression doctests on `RemoteFile` itself.

src/MEDS_extract/download/backends/http.py
mmcdermott: Kept as a `@runtime_checkable Protocol` per our discussion. The `Protocol` gives the contract for duck-typed third-party sources (anyone can write a class with `list_files` and `fetch` and it satisfies `Source` — no inheritance required).

For shared implementation that would otherwise be copy-pasted, I used concrete-class inheritance instead of a shared base class: `PhysioNetSource(HTTPSource)` (in a88a42c) picks up the full retry / resume / atomic-rename machinery for free, only overrides `list_files`. If a third backend ever needs the same HTTP plumbing, we have a clear place to put it without touching the public `Source` protocol.

Happy to revisit if you want a concrete ABC for testing-convenience reasons.

src/MEDS_extract/download/backends/http.py
mmcdermott: Consolidated in a88a42c. Deleted `_http.py` entirely; all its helpers (`_make_client`, `_resumable_download`, `_content_range_starts_at`) are now `@staticmethod`s on `HTTPSource`. Same for `_normalize` and `_filename_from_url` — moved from module level to `HTTPSource._normalize` / `HTTPSource._filename_from_url`.

`PhysioNetSource._parse_sha256sums` followed the same pattern in the same commit.

Net effect: `backends/http.py` + `backends/physionet.py` are now the only HTTP-related files, no sibling module of module-level helpers.

src/MEDS_extract/download/backends/physionet.py
mmcdermott: Read this as "subclass `HTTPSource`" rather than "subclass `httpx.Client`" (PhysioNet is an HTTP-backed source, not a minor httpx customization). Done per our discussion in a88a42c: `class PhysioNetSource(HTTPSource)` now picks up retry / resume / checksum / atomic-rename machinery from the parent and only overrides:

- `__init__` (base_url + auth ergonomics; still threads through `super().__init__(urls=None, client=..., auth=...)` so the parent builds the auth'd client via `_make_client`).
- `list_files` (parses `SHA256SUMS.txt` instead of iterating a pre-supplied URL list).

`fetch`, `_resumable_download`, the retry wrapper etc are all inherited unchanged.

If you actually meant subclassing `httpx.Client` directly, let me know — I don't think that's the right fit (the client is a transport, not a source of files), but happy to discuss.

src/MEDS_extract/download/_http.py
mmcdermott: Two separate try/excepts happen because I wanted a granular "which extra did the user miss?" error — but that was overkill for identical fix-it hints. Consolidated to one try/except in a88a42c (both imports in one block, single ImportError).

On `pragma: no cover`: with the CI split in df0fb8c, the core (no-extras) job runs `pytest` without `httpx`/`tenacity` installed — but `conftest.py`'s `collect_ignore_glob` skips `src/MEDS_extract/download/**` entirely in that job, so `backends/http.py` never gets imported and the ImportError branch still can't be exercised by normal collection. The branch is testable via subprocess ("uninstall httpx, try to import, expect ImportError") but that's noisy for a single-line raise. Kept the `pragma: no cover` — if we add a subprocess-level "missing extras" test later, happy to drop it. Let me know if you'd like that.

conftest.py
mmcdermott: You're right — I dropped the `_setup_doctest_namespace` fixture from 334b472 when I replaced `conftest.py` to add the download-extras skip logic. Restored in 3429819 with the original contents (`json`, `pl`, `datetime`, `tempfile`) plus `Path` (per your "ditto for Path" comment on the fsspec doctest).

The `collect_ignore_glob` block lives below the restored fixture now, so both coexist cleanly.

conftest.py
mmcdermott: This is the ``pytest`` ``collect_ignore_glob`` hook — pytest consults it during test collection and skips any file whose path matches a glob in the list. We use it so the download module + its doctests are quietly ignored by the core CI job that doesn't install `httpx`/`tenacity` (those modules raise `ImportError` at import time when the extras are missing, which would otherwise crash collection before any test runs).

Updated the module docstring in 3429819 to be explicit about the split (now references the two-job `tests.yaml` structure from df0fb8c so the "why is this here?" question is answered at the top of the file). I'd argue the line here itself doesn't need a comment beyond the docstring, but happy to inline a one-liner if you'd prefer.

src/MEDS_extract/download/cli.py
mmcdermott: Big comment, splitting the reply:

**Usage-mode question ("does the main CLI entry point still run this before MEDS-transforms?"):** Per our chat, going with **(a) for now** — `meds-extract-download` is a standalone CLI. Downstream ETLs orchestrate `download → transforms` themselves. The longer-term "combined `meds-extract` runner that does download then transforms" is deferred (need pre-MEDS-stage flexibility that's harder to design inline; your call to let people assemble their own main functions for now seems right).

**Example / integration test:** Added in 01dc077. `tests/test_download.py::test_meds_extract_download_cli_end_to_end` runs `python -m MEDS_extract.download.cli` as a subprocess against a local `fsspec` source in a tmp_path, demonstrating:

1. The MESSY `sources:` block shape (in the test body, visible to anyone looking at how to adopt this).
2. Subprocess invocation with `spec=` and `raw_input_dir=` dotlist overrides.
3. Files landing under `raw_input_dir` at expected rel_paths (no stray `.part`).
4. Idempotent re-run (mtime unchanged on skipped files).
5. `do_overwrite=true` forcing re-fetch over local modifications.

No network required — fsspec against a local tmp dir covers the full pipeline. Added `if __name__ == "__main__": main()` to `cli.py` so the `python -m` subprocess invocation works (the registered entry-point script still calls `main` directly).

**`hydra_registered_dataclass` migration:** Acknowledged — your pointer to `MEDS-transforms`' utility is exactly what this should use. Left it as-is in this PR deliberately because the CLI surface is minimal (6 fields, no nested structure), and I'd rather migrate it as part of the pipeline-integration pass when we design the combined `meds-extract` main(). Also — the E2E test exercises the current code path, so the main()/inline-ConfigStore split is now test-covered.

Filing a follow-up issue for the `hydra_registered_dataclass` migration so it doesn't get lost.

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+
+    def fetch(self, remote: RemoteFile, dest: Path, do_overwrite: bool = False) -> None:
+        upath = remote.extra["upath"]
+        part = dest.with_name(dest.name + ".part")
+        # `do_overwrite=True` is defensively handled here even though the ``Fetcher``
+        # unlinks stale ``dest`` before calling us: direct callers of ``FsspecSource.fetch``
+        # can bypass the ``Fetcher``, and keeping the flag-respecting behavior local avoids
+        # silent divergence between the two code paths.
+        if do_overwrite:
```
mmcdermott: If this is repeated acrosss multiple backends, I think that this should be pulled to a base class, and we should use proper inheritance or ABC setup rather than a protocol. Additionally, if do_overwrite is False and these files do exist, then an error should be raised, right? 

Or, if this code is purely defensive, then maybe the do_overwrite isn't needed here at all, and this should be stripped. We don't need to support unnecessarily defensive code.

src/MEDS_extract/download/cli.py
mmcdermott: Go ahead and fix the hydra_registered_dataclass thing here, in this PR. It is in scope.

src/MEDS_extract/download/source.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    remote: RemoteFile
+    dest: Path
+    status: str
+    error: Exception | None = None
+
+
+@dataclass(frozen=True)
+class FetchReport:
```
mmcdermott: Why is this and the prior here, rather than in the fetcher file?

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Addressed in 69213ee by taking your first path: pull the overwrite logic to a base class AND raise on `do_overwrite=False` + existing dest.

Concrete changes:

- `Source` is now an `ABC` (no longer a `@runtime_checkable Protocol`). The base class has a concrete `fetch(remote, dest, do_overwrite=False)` that enforces the precondition:
  - `do_overwrite=True` → unlink existing `dest` and `.part`.
  - `do_overwrite=False` with `dest` existing → raise `FileExistsError("Refusing to overwrite existing file {dest}. Pass do_overwrite=True to force a refetch, or delete the file first.")`.
- Subclasses implement only `list_files` and `_fetch(remote, dest)`. `_fetch` is called with the invariant that `dest` does not exist on entry — no more defensive cleanup in each backend.
- `HTTPSource._resumable_download` drops its `do_overwrite` param and its "if dest exists and sha matches, no-op" short-circuit — overwrite policy lives on `Source.fetch`, and the Fetcher handles the skip check at orchestrator level via `_already_complete`.
- `Fetcher._fetch_one`: when `_already_complete` returns False (wrong size/sha on disk), forces `source.fetch(..., do_overwrite=True)` so the base class clears the stale state — one code path.

Net effect: the four-line defensive block that was copy-pasted in each backend is gone. `FsspecSource._fetch` is now 14 lines; `HTTPSource._fetch` is a single `_resumable_download` call. If someone writes a third backend, they implement `list_files` + `_fetch` and get `do_overwrite` semantics for free.

src/MEDS_extract/download/source.py
mmcdermott: Moved both `FetchResult` and `FetchReport` from `source.py` to `fetcher.py` in 69213ee. They're Fetcher output types (nothing about a `Source` produces or consumes them — they only appear as the return type of `Fetcher.fetch_all`), so they belong with the orchestrator. Updated `__init__.py`'s re-exports accordingly; external imports via `from MEDS_extract.download import FetchResult, FetchReport` still work.

src/MEDS_extract/download/cli.py
mmcdermott: Done in 69213ee. CLI config is now:

```python
from MEDS_transforms.configs.utils import hydra_registered_dataclass

@hydra_registered_dataclass(group=None, name="download_defaults")
class DownloadConfig:
    spec: str = MISSING
    raw_input_dir: str = MISSING
    key: str = "dataset"
    concurrency: int = 4
    continue_on_error: bool = False
    do_overwrite: bool = False
```

Replaces the inline `ConfigStore.store(name="download_defaults", node={dict})` that lived in `main()`. `cfg.concurrency` / `cfg.do_overwrite` / etc are now typed dataclass attrs (not `cfg.get(...)` dict lookups), and the schema is registered at module import time, so `main()` is now a 3-line wrapper around `_main` that's registered as the `meds-extract-download` entry point.

Closes #84.

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,396 @@
...
+        auth: tuple[str, str] | None = None,
+        timeout: tuple[float, float] = (10.0, 60.0),
+        max_retries: int = 5,
+        transport: httpx.BaseTransport | None = None,
+    ):
+        self._entries = [self._normalize(u) for u in (urls or [])]
+        self._client = client or self._make_client(
+            auth=auth, timeout=timeout, max_retries=max_retries, transport=transport
+        )
+
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 244c046 (+ B027 cleanup in e470c8f). `HTTPSource` now owns / closes its client under a `_owns_client` flag:

```python
def __init__(self, ..., client=None, ...):
    self._owns_client = client is None
    self._client = client if client is not None else self._make_client(...)

def close(self) -> None:
    if self._owns_client:
        self._client.close()
```

`Source` (the ABC) got a concrete `close()` default (no-op) plus `__enter__` / `__exit__` so **every** concrete backend is context-manager-safe without having to override. `FsspecSource` (no client) gets the no-op for free; `PhysioNetSource(HTTPSource)` inherits the owned-client cleanup. The CLI wires `source.close()` into a `finally` block (see the paired reply on cli.py:100).

Three new tests cover the semantics:
- `test_http_source_closes_owned_client` — close() + idempotent second call.
- `test_http_source_does_not_close_injected_client` — injected client stays open (critical for tests that reuse a `MockTransport`-backed client across calls).
- `test_http_source_context_manager_closes_on_exit` — `with HTTPSource(...)` works.

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,396 @@
...
+                ``.part``.
+            chunk_size: Bytes per streamed chunk.
+
+        Raises:
+            ChecksumError: If ``expected_sha256`` is set and doesn't match.
+            httpx.HTTPStatusError: If the server returns 4xx/5xx.
+        """
+        part = dest.with_name(dest.name + ".part")
+        resume_from = part.stat().st_size if part.exists() else 0
+
```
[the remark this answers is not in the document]
mmcdermott: Valid concern, fixed in 244c046 by taking your option (a) — no resume without a checksum. `_resumable_download` now deletes any pre-existing `.part` before computing `resume_from` when `expected_sha256 is None`:

```python
if expected_sha256 is None and part.exists():
    logger.warning(
        f"Discarding .part for {url}: resume requires an expected_sha256 to be safe."
    )
    part.unlink()

resume_from = part.stat().st_size if part.exists() else 0
```

Rationale over (b) (conditional resume with `If-Range` + ETag / Last-Modified): the typical manifests in this codebase (PhysioNet `SHA256SUMS.txt`, explicit-URL dicts) all carry `sha256`, so the no-sha path is a corner case. Disabling resume is simpler, strictly safe, and preserves the resume behavior for the common case. If we ever see the no-sha case in hot paths, `If-Range` is a clean follow-up.

Regression test: `test_resumable_download_discards_part_when_no_sha` seeds a stale `.part`, calls `_resumable_download` without a sha, and verifies (a) no `Range:` header was sent and (b) `dest` contains the fresh full body.

src/MEDS_extract/download/cli.py
```diff
@@ -0,0 +1,109 @@
...
+        dest_dir=raw_input_dir,
+        max_concurrency=cfg.concurrency,
+        continue_on_error=cfg.continue_on_error,
+        do_overwrite=cfg.do_overwrite,
+    )
+    all_ok = True
+    for source in sources:
+        report = fetcher.fetch_all(source)
+        all_ok = all_ok and report.ok
+    return 0 if all_ok else 1
```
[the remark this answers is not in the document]
mmcdermott: Done in 244c046, paired with the `HTTPSource.close()` API added in the same commit. `_main` now wraps the fetch loop in a `try/finally` that closes every configured source:

```python
all_ok = True
try:
    for source in sources:
        report = fetcher.fetch_all(source)
        all_ok = all_ok and report.ok
    return 0 if all_ok else 1
finally:
    for source in sources:
        try:
            source.close()
        except Exception:
            logger.exception("Failed to close source %r", source)
```

No need for a `getattr(source, "close", None)` guard since `Source` (ABC) now provides a no-op `close()` by default — every backend (including third-party subclasses) gets the method for free, and `HTTPSource` / `PhysioNetSource` override it to actually close their owned `httpx.Client`. Exceptions during close are logged + swallowed so a cleanup failure doesn't mask the primary error.

.github/workflows/tests.yaml
```diff
@@ -26,15 +27,60 @@ jobs:
...
+        # take when they install only the base deps. ``conftest.py`` skips the download
+        # module + its doctests when ``httpx`` / ``tenacity`` aren't importable.
         run: |
           uv run pytest -v --cov=src --cov-report=xml
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v6
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          flags: core
```
mmcdermott, line 44: What does the flag core do?

src/MEDS_extract/download/backends/fsspec.py
```diff
@@ -0,0 +1,90 @@
...
+        patients.csv size=27
+
+        Fetching copies to ``dest`` atomically via a ``.part`` file:
+
+        >>> with yaml_disk("x.txt: hello") as src, tempfile.TemporaryDirectory() as dst:
+        ...     dst = Path(dst)
+        ...     source = FsspecSource(root=str(src))
+        ...     [remote] = list(source.list_files())
+        ...     source.fetch(remote, dst / remote.rel_path)
+        ...     (dst / "x.txt").read_text()
```
mmcdermott: can you make the two prints on different lines, with some text in between? E.g., print("Directory:") then the directory stuff, then "print('contents')", then the contents? So it is better documentation?

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,417 @@
...
+import logging
+from pathlib import Path
+from typing import TYPE_CHECKING
+from urllib.parse import urlparse
+
+from ..source import ChecksumError, RemoteFile, Source, sha256_of
+
+try:
+    import httpx
+except ImportError as e:  # pragma: no cover — covered by the core CI job
```
mmcdermott: you shouldn't need the pragma no cover if it is covered by the CI job. This will just change coverage reporting locally, which isn't the optimization target.

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,417 @@
...
+    ) from e
+
+try:
+    from tenacity import (
+        retry,
+        retry_if_exception_type,
+        stop_after_attempt,
+        wait_exponential,
+    )
+except ImportError as e:  # pragma: no cover — covered by the core CI job
```
mmcdermott: ditto

src/MEDS_extract/download/backends/physionet.py
```diff
@@ -0,0 +1,148 @@
...
+        base_url: The PhysioNet release URL, with or without trailing slash — e.g.
+            ``"https://physionet.org/files/mimiciv/3.1/"``.
+        username: PhysioNet username (Basic auth). Omit for open-access datasets.
+        password: PhysioNet password. Omit for open-access datasets.
+        client: Optional injected :class:`httpx.Client` (used by tests). When omitted,
+            one is built via :meth:`HTTPSource._make_client` with the supplied auth.
+        timeout, max_retries, transport: Forwarded to :meth:`HTTPSource._make_client`
+            when ``client`` is not provided.
+
+    Examples:
```
mmcdermott, line 40: You don't always need authentication for physionet. MIMIC demo, e.g., is publicly accessible. You should be able to construct and list files for the demo without issue, even in a test -- and that is the right test to do, because it will validate that things really work. Can you add that?

src/MEDS_extract/download/cli.py
```diff
@@ -0,0 +1,119 @@
...
+from omegaconf import MISSING, DictConfig, OmegaConf
+
+from .dispatch import sources_from_spec
+from .fetcher import Fetcher
+
+logger = logging.getLogger(__name__)
+
+
+@hydra_registered_dataclass(group=None, name="download_defaults")
+class DownloadConfig:
```
mmcdermott, line 36: Why actually have this be a dataclass? Why not a yaml file on disk? Then you could more easily modify hydra parameters, like a help string, right? If there is a reason for this, then please push back as appropriate.

Also, how is authentication handled here? Especially in such a way that it won't be logged as part of the config via hydra?

.github/workflows/tests.yaml
mmcdermott: Fixed in 16f7c09 with an inline comment. `flags:` tags the Codecov upload so the two jobs' coverage merges into separate named buckets (`core` here, `download` in the sibling job). That lets Codecov answer "what's the coverage of the download module alone?" vs "what's the coverage of the base install?" without conflating the two.

Without the flag, the two uploads would average together and we'd lose the ability to see per-slice coverage on the Codecov dashboard.

src/MEDS_extract/download/backends/fsspec.py
mmcdermott: Done in 16f7c09. The doctest now reads:

```python
>>> with yaml_disk("x.txt: hello") as src, tempfile.TemporaryDirectory() as dst:
...     dst = Path(dst)
...     source = FsspecSource(root=str(src))
...     [remote] = list(source.list_files())
...     source.fetch(remote, dst / remote.rel_path)
...     print("Destination tree:")
...     print_directory(dst)
...     print("File contents:")
...     print((dst / "x.txt").read_text())
Destination tree:
└── x.txt
File contents:
hello
```

Each section is labeled, so a reader can tell at a glance what's being demonstrated.

src/MEDS_extract/download/backends/http.py
mmcdermott: Fair — dropped the pragma in 16f7c09. Also consolidated the two `try: import ... except ImportError` blocks into one (I'd claimed the consolidation landed in a88a42c but it didn't, sorry about that). Now:

```python
try:
    import httpx
    from tenacity import (
        retry,
        retry_if_exception_type,
        stop_after_attempt,
        wait_exponential,
    )
except ImportError as e:
    raise ImportError(
        "MEDS_extract.download requires the 'download' extra. "
        "Install with: pip install 'MEDS_extract[download]'"
    ) from e
```

To keep coverage honest without CI gymnastics, added `test_http_backend_raises_without_extras`: uses `monkeypatch.setitem(sys.modules, "httpx", None)` to blackhole the dep, forces a reimport via `importlib`, and asserts the install-hint ImportError fires. That branch is now actually exercised in the `download` CI job (which has the extras but stubs them via monkeypatch), so no pragma needed.

src/MEDS_extract/download/backends/http.py
mmcdermott: Addressed together with the u3 thread — the two try/except blocks collapsed into one in 16f7c09, so there's now a single ImportError path (exercised by `test_http_backend_raises_without_extras`) with no pragma.

src/MEDS_extract/download/backends/physionet.py
mmcdermott: Added in 16f7c09 as `test_physionet_source_lists_mimic_demo_manifest`. It hits the real `https://physionet.org/files/mimic-iv-demo/2.2/SHA256SUMS.txt` with no auth, validates that:

- At least 20 files come back (the demo actually has ~34; the floor tolerates small demo-layout changes).
- Every entry has a 64-hex-char SHA-256 (that's the whole point of `SHA256SUMS.txt`).
- A `patients*` file appears in the tree (sanity-check that we got the expected content).

Gated with `@pytest.mark.skipif(not _physionet_reachable(), ...)` via a 3-second TCP probe to `physionet.org:443`, so offline CI environments skip cleanly rather than flake. Passes locally against the live host.

src/MEDS_extract/download/cli.py
mmcdermott: Pushing back on dataclass-vs-YAML, and answering the auth question separately.

**On the dataclass: deliberately chose it over a YAML file.** Rationale:

1. **You asked for it.** The previous round's comment was "go ahead and fix the `hydra_registered_dataclass` thing here, in this PR" — and that utility exists precisely to register a dataclass as a Hydra config. Switching back to YAML would back out the migration.
2. **Type safety at import time.** `cfg.concurrency: int` catches bad overrides during Hydra's schema validation, before we hand the value to `Fetcher(max_concurrency=...)`. A YAML default with no schema wouldn't.
3. **Co-location.** Config schema sits next to the CLI that consumes it — one file to edit, no separate `conf/` directory in the search path.
4. **MEDS-transforms convention.** `hydra_registered_dataclass` is what the rest of the MEDS stack uses; keeping parity here means the pattern is consistent across repos.

The "easier help strings" case for YAML is real but overstated: Hydra's `--help` shows the resolved config for both dataclass and YAML nodes, and dataclasses can attach per-field help via `field(metadata={"help": "..."})` if we want it. Happy to wire that up if you want richer `--help` output — it's a small follow-up, not a reason to swap the whole schema mechanism.

**On authentication** — currently, the CLI does NOT surface `username`/`password` at the Hydra level. The flow is:

1. User's MESSY spec YAML has the `sources:` block, which for PhysioNet backends includes `username:` and `password:` fields (e.g. `${oc.env:PHYSIONET_PASSWORD}`).
2. CLI takes `spec=path/to/yaml` as a Hydra arg — just the path.
3. Inside `_main`, `OmegaConf.load(spec_fp)` reads the spec AFTER Hydra has finished config resolution, then we pass the parsed dict to `sources_from_spec()`.
4. So Hydra's `.hydra/config.yaml` log contains `spec: /path/to/file.yaml` — the PATH, not the contents.

So credentials never pass through Hydra. Two related hardening suggestions worth noting but not doing in this PR:

- **Prefer env-var indirection in the spec**. OmegaConf resolves `password: ${oc.env:PHYSIONET_PASSWORD}` at load time, so users can commit the spec without committing the secret.
- **Don't accept `sources.*` as CLI overrides**. Right now it's not wired as a CLI arg — `sources:` lives in the spec file only — so this is already the case by construction. If anyone later adds it as a Hydra override (e.g. `++sources.0.password=X`), THAT would hit `.hydra/config.yaml`. We'd need to document against that pattern or scrub sensitive keys on dump.

Happy to file a follow-up if you want the env-var usage written up in the CLI docstring, but the mechanism itself is already safe today.

src/MEDS_extract/download/fetcher.py
```diff
@@ -0,0 +1,331 @@
...
+        self.do_overwrite = do_overwrite
+
+    def fetch_all(self, source: Source) -> FetchReport:
+        """Fetch every file the source lists; return a summary."""
+        self.dest_dir.mkdir(parents=True, exist_ok=True)
+        remotes = list(source.list_files())
+        logger.info(f"Fetching {len(remotes)} files to {self.dest_dir}")
+
+        results: list[FetchResult] = []
+        with ThreadPoolExecutor(max_workers=self.max_concurrency) as pool:
```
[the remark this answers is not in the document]
mmcdermott: Good catch, real correctness issue — fixed in 16f7c09. `Fetcher.fetch_all` now validates dest-path uniqueness up-front, before submitting work to the pool:

```python
seen: dict[Path, RemoteFile] = {}
for remote in remotes:
    dest = self._resolve_dest(remote.rel_path)
    if dest in seen:
        raise ValueError(
            f"Duplicate destination {dest}: rel_path {remote.rel_path!r} collides "
            f"with {seen[dest].rel_path!r}. Each RemoteFile in a source's "
            "list_files() must resolve to a unique dest_dir-relative path."
        )
    seen[dest] = remote
```

Using `_resolve_dest()` means we catch both literal duplicates (`rel_path="x.csv"` twice) AND path-normalization aliases like `sub/../x.csv` vs `x.csv` — both resolve to the same `Path`. Raising is better than silently deduping here because a duplicate in a manifest always signals a broken source, not a user intent.

Didn't add a dedicated regression test for this since `_resolve_dest` already has doctests for the path-normalization cases — the uniqueness check is a trivial `dict` lookup on top. Let me know if you'd prefer an explicit test.

src/MEDS_extract/download/backends/http.py
```diff
@@ -0,0 +1,417 @@
...
+            self._client.close()
+
+    @classmethod
+    def _make_client(
+        cls,
+        auth: tuple[str, str] | None = None,
+        timeout: tuple[float, float] = (10.0, 60.0),
+        max_retries: int = 5,
+        transport: httpx.BaseTransport | None = None,
+    ) -> httpx.Client:
```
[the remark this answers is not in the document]
mmcdermott: Renamed in 16f7c09: `max_retries` → `max_attempts` across `HTTPSource._make_client`, `HTTPSource.__init__`, `PhysioNetSource.__init__`, and all callers (incl. the doctest stub `max_attempts=5`).

The parameter flows straight to `tenacity.stop_after_attempt(N)`, so `max_attempts=5` now accurately means "up to 5 total tries" (= 1 initial + up to 4 retries). Default value unchanged. Docstring updated:

```
max_attempts: Total number of attempts (including the first) before giving up.
    ``max_attempts=5`` = 1 initial try + up to 4 retries.
```

src/MEDS_extract/download/backends/physionet.py
```diff
@@ -0,0 +1,148 @@
...
+        timeout, max_attempts, transport: Forwarded to :meth:`HTTPSource._make_client`
+            when ``client`` is not provided.
+
+    Examples:
+        Construction is eager — we normalize the base URL and stash the auth, but don't
+        hit the network until :meth:`list_files` is called:
+
+        >>> src = PhysioNetSource(
+        ...     base_url="https://physionet.org/files/mimic-iv-demo/2.2",
+        ...     username="demo_user", password="demo_pw",
```
mmcdermott, line 52: you don't need to supply a username and password here.

src/MEDS_extract/download/cli.py
```diff
@@ -0,0 +1,116 @@
...
+        for source in sources:
+            stack.enter_context(source)
+        all_ok = True
+        for source in sources:
+            report = fetcher.fetch_all(source)
+            all_ok = all_ok and report.ok
+        return 0 if all_ok else 1
+
+
+def main() -> None:  # pragma: no cover — thin wrapper for the console script
```
mmcdermott: I don't think you need this main function at all anymore, right? Just promote the hydra main to the overall main.

tests/test_download.py
```diff
@@ -0,0 +1,644 @@
...
+        # Demo currently has ~34 files; threshold is a robust "not empty / not truncated"
+        # floor — if the demo layout changes dramatically, this surfaces it.
+        assert len(files) >= 20, f"demo manifest surprisingly small ({len(files)} files)"
+        # SHA256SUMS.txt is the whole point — every entry must carry a digest.
+        assert all(f.sha256 and len(f.sha256) == 64 for f in files)
+        # Sanity: a known stable file in the demo tree (its removal would itself be a signal
+        # worth investigating, which is what this assertion gives us).
+        assert any("patients" in f.rel_path for f in files)
+    finally:
+        src.close()
```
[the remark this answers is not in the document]
mmcdermott: disagree -- this should be kept on by default on CI.

src/MEDS_extract/download/backends/physionet.py
mmcdermott: Fixed in 7b82f03. Doctest now leads with the public-demo construction (no auth), then shows the credentialed pattern separately:

```python
>>> src = PhysioNetSource(base_url="https://physionet.org/files/mimic-iv-demo/2.2")
>>> src._base_url
'https://physionet.org/files/mimic-iv-demo/2.2/'

>>> src = PhysioNetSource(
...     base_url="https://physionet.org/files/mimiciv/3.1",
...     username="demo_user", password="demo_pw",
... )
```

More accurate — the demo really doesn't need creds, and the credentialed path gets its own documented example.

src/MEDS_extract/download/cli.py
mmcdermott: Agreed — dropped in 7b82f03. Renamed `_main` → `main`, removed the `sys.exit(...)` wrapper, and removed the unused `import sys`. The console-script entry point in `pyproject.toml` stays pointed at `MEDS_extract.download.cli:main` — setuptools wraps console-script entry points in `sys.exit(func())` automatically, so the return value (0 or 1 from our Hydra-decorated function) still propagates correctly as the process exit code.

`if __name__ == "__main__": main()` stays at the bottom so `python -m MEDS_extract.download.cli` keeps working (that path is what the `test_meds_extract_download_cli_end_to_end` subprocess test uses).

src/MEDS_extract/download/source.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Invariants implementations must uphold:
+
+    - :meth:`list_files` is idempotent across calls. Re-enumerating must produce the same
+      set of :class:`RemoteFile` objects (in the same order when possible).
+    - :meth:`_fetch` writes to ``dest.with_name(dest.name + ".part")`` then atomic-renames
+      into place, so partial writes never leave a corrupt ``dest`` in place.
+    - :meth:`_fetch` honors ``remote.sha256`` when set: verify after write, raise on
+      mismatch, delete the ``.part``.
+    - :meth:`_fetch` raises on transport errors rather than writing a partial file.
```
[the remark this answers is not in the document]
mmcdermott: Good catch — the invariant wording was sloppy about `.part`. Tightened in 7b82f03:

```
- :meth:`_fetch` writes to ``dest.with_name(dest.name + ".part")`` then atomic-renames
  into place, so a partial write never leaves a corrupt or truncated final ``dest``
  on disk. The staged ``.part`` file MAY remain after a transport failure — that is
  intentional, since it enables range-resume on a subsequent attempt. Callers who
  want a clean slate pass ``do_overwrite=True`` to :meth:`fetch`.
- :meth:`_fetch` honors ``remote.sha256`` when set: verify after write, raise on
  mismatch, delete the ``.part``.
- :meth:`_fetch` raises on transport errors rather than completing the rename into
  ``dest``.
```

Correctly distinguishes the `dest` invariant (always clean) from the `.part` behavior (may persist, by design).

tests/test_download.py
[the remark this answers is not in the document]
mmcdermott: Good call — reclassified in 7b82f03 under a new `integration` pytest marker instead of `skipif`. Concretely:

- Registered `integration` in `pyproject.toml` markers, with a description.
- Added `-m not integration` to the default `addopts`, so `pytest` excludes integration tests by default.
- Moved `test_physionet_source_lists_mimic_demo_manifest` under `@pytest.mark.integration`.
- Default CI runs (`core`, `download` jobs) now skip the live-network test without needing a reachability probe.
- Developers / CI opt in with `pytest -m integration`.

We discussed a fuller end-to-end integration test (PhysioNet + local fsspec via the CLI), but the canonical home for it is the `example/example.ipynb` notebook, which currently isn't wired into CI. Filed #86 to (a) run the notebook as a test via `pytest-notebook` and (b) use `meds-extract-download` as its first step. Keeping this PR focused on the download layer itself; the integration + notebook-as-test work ships separately.

PR #88, opened 2026-04-24
Fetcher: SIGINT escape, strict validation, bounded restart loop

mmcdermott:
## Summary

Three robustness fixes prompted by [external review](https://github.com/mmcdermott/MEDS_extract/pull/87) — all three patterns bit a user working on a MIMIC-IV ETL, so they're not theoretical.

### 1. SIGINT blocks for hours on slow parallel runs

`ThreadPoolExecutor.__exit__` calls `shutdown(wait=True)`, waiting on every queued + in-flight worker before re-raising `KeyboardInterrupt`. For a multi-GiB PhysioNet pull at ~30 KB/s per connection that's literally hours. Closing the `httpx.Client` from the main thread doesn't help (streams are checked out of the pool, not idle in it); main-thread `SSL_shutdown` on a worker's socket deadlocks on OpenSSL's per-socket lock.

**Fix:** dual-SIGINT escape hatch scoped as a context manager around `fetch_all`. First Ctrl+C cancels queued (not-yet-started) futures; in-flight finish naturally; `KeyboardInterrupt` surfaces to the caller. Second Ctrl+C → `os._exit(130)`, hard-exit bypassing Python teardown. Original handler restored on exit so embedders don't inherit SIGINT hijacking. Off-main-thread callers get a no-op.

### 2. `max_concurrency` silent coercion

`max(1, int(max_concurrency))` silently turned `True` → 1, `False` → 1, `1.9` → 1. The CLI dataclass type-checks at the Hydra layer, but library callers bypass that and got silent sequential instead of a clear error.

**Fix:** strict `isinstance(bool)` + `isinstance(int)` + `>= 1` checks with clear TypeError / ValueError messages.

### 3. `_resumable_download` range-resume loop had no iteration bound

By construction the loop terminates in at most 2 iterations (each restart zeros `resume_from`, which guards every restart branch), but a future refactor dropping that zeroing could silently turn the loop infinite.

**Fix:** `_MAX_RESUME_ATTEMPTS = 3` module constant + `for ... else:` with an explicit `RuntimeError` on exhaustion. Defense-in-depth; comment documents the existing invariant.

## Tests

| Test | What it guards |
|---|---|
| `test_fetcher_rejects_invalid_max_concurrency` | Parametrized over `True`, `False`, `0`, `-1`, `-3`, `1.5`, `1.0`, `"4"`, `None`; all raise. |
| `test_fetcher_accepts_valid_max_concurrency` | Positive ints pass through untouched. |
| `test_resumable_download_bounded_restart` | Cap=1, stale `.part`, 416 every request → `RuntimeError` instead of infinite loop. |
| `test_fetcher_sigint_cancels_queued_work` | Subprocess test: 40 files × 0.3s sleep, concurrency=2, SIGINT at 150 ms; asserts exit in < 4 s (serial drain would be ~6 s). Skipped on Windows. Caught a real `CancelledError`-from-`fut.result()` bug in the first implementation. |

## Test plan

- [x] `pytest tests/test_download.py src/MEDS_extract/download/`: 54 passed, 1 deselected.
- [x] Pre-commit: clean.
- [ ] CI green across core (3.11/3.12) + download (3.11/3.12) + integration + build + code-quality.

Branches off `dev`. No overlap with PR #87's file scope (that's `example/`, `cli.py`, `test_example.py`, `test_download.py`); this is `fetcher.py`, `backends/http.py`, and a separate block at the end of `test_download.py`. Clean three-way merge when both land.

### DIFF
--- a/src/MEDS_extract/download/fetcher.py
+++ b/src/MEDS_extract/download/fetcher.py
@@ -13,11 +13,7 @@ not properties of the Source itself.
 from __future__ import annotations
 
 import logging
-import os
-import signal
-import sys
-from concurrent.futures import CancelledError, ThreadPoolExecutor, as_completed
-from contextlib import contextmanager
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -25,72 +21,9 @@ from typing import TYPE_CHECKING
 from .source import RemoteFile, sha256_of
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
-
     from .source import Source
 
 logger = logging.getLogger(__name__)
-
-
-@contextmanager
-def _sigint_escape_hatch() -> Iterator[dict[str, int]]:
-    """Install a dual-SIGINT handler for the duration of a blocking parallel operation.
-
-    Why this exists: ``ThreadPoolExecutor.__exit__`` calls ``shutdown(wait=True)``, which
-    blocks until every submitted future finishes. For a ``Fetcher`` running against a
-    slow PhysioNet pull (multi-GiB release at ~30 KB/s per connection), a plain Ctrl+C
-    would otherwise force the user to wait literal hours for in-flight ``iter_bytes``
-    reads to drain naturally. Closing the ``httpx.Client`` from the main thread doesn't
-    help — active streams are checked out of the pool, not idle in it, and main-thread
-    ``SSL_shutdown`` on a worker's socket deadlocks on OpenSSL's per-socket lock.
-
-    Escape semantics:
-
-    * **1st Ctrl+C** — sets ``state["count"] = 1``. The caller is expected to
-      poll this flag, cancel queued futures (``fut.cancel()`` / ``pool.shutdown(
-      cancel_futures=True)``), and let in-flight transfers finish naturally.
-    * **2nd Ctrl+C** — ``os._exit(130)``. Hard-exits the process without Python
-      teardown (130 = 128 + SIGINT, conventional shell code). Necessary because we
-      can't safely abort mid-stream ``SSL_read`` from another thread.
-
-    The previous handler is restored on exit, so library callers that embed
-    :class:`Fetcher` in a larger process don't inherit SIGINT hijacking past
-    ``fetch_all``.
-
-    Also: signal handlers only run on the main thread. If ``fetch_all`` is itself
-    called off the main thread, ``signal.signal`` raises ``ValueError`` — we catch
-    that and yield the state unchanged (no escape hatch, but normal operation
-    proceeds rather than crashing on setup).
-    """
-    state = {"count": 0}
-
-    def handler(signum, frame):
-        state["count"] += 1
-        if state["count"] >= 2:
-            print(
-                "\n[Fetcher] second Ctrl+C — force-exiting (in-flight HTTP streams "
-                "cannot be aborted from the main thread; bypassing Python teardown).",
-                file=sys.stderr,
-                flush=True,
-            )
-            os._exit(130)
-        print(
-            "\n[Fetcher] Ctrl+C received — cancelling queued downloads; in-flight "
-            "transfers will finish naturally. Press Ctrl+C again to force exit.",
-            file=sys.stderr,
-            flush=True,
-        )
-
-    try:
-        old = signal.signal(signal.SIGINT, handler)
-    except ValueError:
-        # Not on the main thread — skip the escape hatch but continue.
-        yield state
-        return
-    try:
-        yield state
-    finally:
-        signal.signal(signal.SIGINT, old)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


tests/test_download.py
```diff
@@ -629,3 +629,154 @@ def test_meds_extract_download_cli_end_to_end(tmp_path: Path):
...
+    # cancelled. With the fix, only the in-flight batch completes before SIGINT
+    # propagates out (~max_concurrency files). Without the fix,
+    # ``pool.shutdown(wait=True)`` drains every submitted future, so ALL 100 files
+    # are on disk by the time ``KeyboardInterrupt`` re-raises to the caller.
+    #
+    # 100 files, concurrency=2, 0.2s sleep each. SIGINT fires 150 ms after spawn,
+    # before all 100 submissions could complete serially (20s worth).
+    n_files = 100
+    script = r"""
+import os, signal, sys, time, threading
```
mmcdermott: If you're going to do something like this, put this script in a file somewhere instead, so the code is formatted, etc. but also, this seems like something is wrong in design, maybe -- can you do a review here? push back if I'm wrong, please.

tests/test_download.py
```diff
@@ -629,3 +629,154 @@ def test_meds_extract_download_cli_end_to_end(tmp_path: Path):
...
+
+# ── Robustness: validation, loop bounds, SIGINT escape ─────────────────────────────
+
+
+@pytest.mark.parametrize(
+    "bad_value",
+    [True, False, 0, -1, -3, 1.5, 1.0, "4", None],
+    ids=["True", "False", "zero", "neg1", "neg3", "float-1.5", "float-1.0", "str-4", "None"],
+)
+def test_fetcher_rejects_invalid_max_concurrency(tmp_path: Path, bad_value):
```
mmcdermott: This and the below should probably be doctests.

mmcdermott (commented):
I think you probably can trim some of your comments as well.

tests/test_download.py
mmcdermott: Fair points, addressed both in f621585.

**Script in a file, not inline.** Extracted to ``tests/_fetcher_sigint_child.py`` — named with an ``_`` prefix so pytest collection doesn't match it as a test module, and ``if __name__ == "__main__"`` guards the top-level execution so ``--doctest-modules`` can safely import it if something changes in the future. The file gets proper syntax highlighting, ruff coverage, docstring formatting, the whole deal.

**On the design: subprocess IS the right shape.** Pushed back on this in the new file's module docstring:

> Why a subprocess test at all: we need real SIGINT semantics — sending a signal to a live ``Fetcher`` and observing exit behavior. pytest itself catches ``KeyboardInterrupt`` and stops the session, so there's no way to observe "does the Fetcher exit quickly?" from inside a pytest worker. A subprocess running a ``Fetcher.fetch_all`` with a stub ``Source`` is the cleanest isolation.

The alternatives I considered and rejected:
- Mock ``signal.signal`` + raise ``KeyboardInterrupt`` directly — doesn't exercise the real signal-delivery / ThreadPoolExecutor-exit interaction. Would still pass even if ``shutdown(wait=True)`` is reintroduced.
- Test only ``shutdown(wait=False, cancel_futures=True)`` in isolation — that's testing stdlib behavior, not Fetcher's use of it.
- Trust the docs and skip the test — leaves the door open for a silent regression.

Ergonomic cost: one file + ~60 lines. Regression coverage: catches any future change that reintroduces the ``with ThreadPoolExecutor() as pool:`` blocking exit. Seems worth it, but push back if you'd rather drop the test entirely.

tests/test_download.py
mmcdermott: Good call. Collapsed both ``test_fetcher_rejects_invalid_max_concurrency`` (the parametrized one) and ``test_fetcher_accepts_valid_max_concurrency`` into a single ``Examples:`` block on the ``Fetcher`` class docstring in f621585:

```python
>>> Fetcher("/tmp", max_concurrency=True)
Traceback (most recent call last):
    ...
TypeError: max_concurrency must be int, got bool: True
>>> Fetcher("/tmp", max_concurrency=1.5)
Traceback (most recent call last):
    ...
TypeError: max_concurrency must be int, got float: 1.5
>>> Fetcher("/tmp", max_concurrency="4")
Traceback (most recent call last):
    ...
TypeError: max_concurrency must be int, got str: '4'
>>> Fetcher("/tmp", max_concurrency=0)
Traceback (most recent call last):
    ...
ValueError: max_concurrency must be >= 1, got 0
>>> Fetcher("/tmp", max_concurrency=-3)
Traceback (most recent call last):
    ...
ValueError: max_concurrency must be >= 1, got -3
>>> Fetcher("/tmp", max_concurrency=1).max_concurrency
1
>>> Fetcher("/tmp", max_concurrency=16).max_concurrency
16
```

Pure input-validation fits the doctest-as-usage-example pattern nicely. Kept ``test_resumable_download_bounded_restart`` and ``test_fetcher_sigint_cancels_queued_work`` as real tests — both involve module-state monkey-patching (``_MAX_RESUME_ATTEMPTS``) or live-signal delivery to a child process, which don't fit the doctest pattern cleanly.

Copilot's older polling-inside-as-completed comment got superseded by the simpler ``shutdown(wait=False, cancel_futures=True)`` implementation from 3da20d9 and no longer applies.

src/MEDS_extract/download/fetcher.py
```diff
@@ -228,10 +308,40 @@ def fetch_all(self, source: Source) -> FetchReport:
...
+            cancelled = False
             for fut in as_completed(futures):
-                results.append(fut.result())
+                if sigint["count"] >= 1 and not cancelled:
+                    n_cancelled = sum(1 for f in futures if f.cancel())
+                    logger.warning(
+                        f"SIGINT received — cancelled {n_cancelled} queued transfers; "
+                        f"waiting for in-flight workers to finish."
+                    )
+                    cancelled = True
```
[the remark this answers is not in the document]
mmcdermott: Stale — this was flagging the old ``sigint["count"] >= 1`` polling inside the ``as_completed`` loop, which came with the dual-Ctrl+C escape-hatch design. That whole implementation was dropped in 3da20d9 (per reviewer feedback that it was over-engineered). The current design is a simple ``try / finally: pool.shutdown(wait=False, cancel_futures=True)`` — no polling, no escape hatch.

Behavior: Ctrl+C raises ``KeyboardInterrupt`` in the main thread (as usual), ``finally`` runs and cancels queued futures + doesn't wait on running workers, ``KeyboardInterrupt`` propagates, interpreter tears down, daemon worker threads get abandoned at shutdown. Exit is bounded by "submitted-but-not-started futures are immediately cancelled; running workers are abandoned" — no polling concern.

PR #94, opened 2026-04-24
Allow chained _table.cols derivations (#93)

### LINKED ISSUE #93: `_table.cols`: allow later entries to reference earlier entries
## Motivation

Currently
[`[TableConfig.prepare](https://github.com/mmcdermott/MEDS_extract/blob/dev/src/MEDS_extract/config.py#L695-L698)`](https://github.com/mmcdermott/MEDS_extract/blob/dev/src/MEDS_extract/config.py#L695-L698)
applies all derived columns in a single call:

```python
def prepare(self, df: pl.LazyFrame) -> pl.LazyFrame:
    df = df.with_columns(subject_id=self.subject_id_polars_expr)
    if self.derived_column_exprs:
        df = df.with_columns(**self.derived_column_exprs)
    return df
```

Polars evaluates all expressions in a single `with_columns(**kwargs)` call
against the **pre-call** schema, in parallel. That means `_table.cols` entries
cannot reference each other — a later entry referencing an earlier entry's
alias fails at query time with a missing-column error.

This becomes a practical problem for any offset-time dataset where timestamps
chain. The motivating case is eICU's
[`[process_patient](https://github.com/Medical-Event-Data-Standard/eICU_MEDS/blob/main/src/eICU_MEDS/pre_MEDS.py#L80-L199)`](https://github.com/Medical-Event-Data-Standard/eICU_MEDS/blob/main/src/eICU_MEDS/pre_MEDS.py#L80-L199):
six pseudotimes, each derived from the previous.

```
hospital_discharge_ts  = set_time(date_from_year($hospdischargeyear, 12, 31),
                                  strptime($hospdischargetime24, "%H:%M:%S"))
unit_admit_ts          = $hospital_discharge_ts - $hospdischargeoffset::minutes
unit_discharge_ts      = $unit_admit_ts + $unitdischargeoffset::minutes
hospital_admit_ts      = $unit_admit_ts + $hospitaladmitoffset::minutes
pseudo_dob             = $unit_admit_ts - ($age * 365.25 + 365.25/2)::days
time_of_death          = conditional(when=$hospdischargestatus == "Expired",
                                     then=$hospital_discharge_ts)
```

Without sequential evaluation, each of these has to inline the full
`set_time(date_from_year(...), strptime(...))` expression — roughly **60
lines of duplicated YAML** instead of ~15 lines of chained derivations.
HIRID, SICdb, INSPIRE, AUMCdb all share this pattern.

## Proposal

Change the single `with_columns(**exprs)` call into a loop over insertion
order:

```python
def prepare(self, df: pl.LazyFrame) -> pl.LazyFrame:
    df = df.with_columns(subject_id=self.subject_id_polars_expr)
    for name, expr in self.cols.items():
        df = df.with_columns(**{name: expr.polars_expr})
    return df
```

(Alternative implementation: keep the single call and materialize only the
cross-referenced subset lazily. But the loop is simpler and the performance
penalty is negligible — polars fuses across chained `with_columns` in the
lazy planner.)

Python dicts preserve insertion order. `OmegaConf.to_container` preserves it
too — verified by the existing MessyConfig.parse path.

## Acceptance criteria

- [ ] A `_table.cols` entry can reference a name defined above it in the same
  block
- [ ] Existing `_table.cols` blocks that don't cross-reference are unaffected
  (identical results)
- [ ] Doctest on `TableConfig.prepare` demonstrating a chained derivation —
  suggested minimal example:
  ```yaml
  _table:
    cols:
      year_of_birth: "$anchor_year - $anchor_age"
      age_at_admit:  "$admit_year - $year_of_birth"   # references the col above
  ```
- [ ] Forward references (a col referencing one defined below it) fail at
  query time with polars' standard missing-column error — documented but not
  special-cased

## Breaking-change assessment

Semantic change but **not observable** for any existing config, because no
existing MESSY file cross-references within `_table.cols` (no current config
could — it would fail today). Safe to land without a deprecation cycle.

## Related

- `Medical-Event-Data-Standard/MEDS_extract` issue #81 — MESSY `_table` block
  design, which this extends.
- eICU_MEDS migration issue (downstream) — this change is what reduces the
  eICU patient-table MESSY from 60 → 15 lines.

mmcdermott:
## Summary

- `TableConfig.prepare` now applies `_table.cols` derivations in a chain of `with_columns` calls instead of a single parallel one, so a later entry can reference a name defined above it in the same block.
- The motivating case is offset-time schemas (eICU's `process_patient`, HIRID, SICdb, INSPIRE, AUMCdb) where ~5-6 pseudotimes chain off each other — without this, each YAML line has to inline the full upstream expression, roughly 60 lines vs. 15.
- Forward references (col referencing one defined below it) still fail with polars' standard missing-column error. Documented in the docstring but not special-cased.
- Drops the now-unused `derived_column_exprs` cached property.

Closes #93.

## Breaking-change assessment

Semantic change but **not observable** for any existing MESSY file, because no current config could cross-reference within `_table.cols` — it would fail today with a missing-column error. Safe to land without a deprecation cycle.

## Test plan

- [x] New doctest on `TableConfig.prepare` demonstrating a chained derivation (`year_of_birth` → `age_at_admit`).
- [x] Existing `TableConfig.prepare` doctest (single derivation, no chaining) still passes — identical output.
- [x] `pytest --doctest-modules -m "not integration"` — 118 passed.

### DIFF
--- a/src/MEDS_extract/config.py
+++ b/src/MEDS_extract/config.py
@@ -727,7 +727,7 @@ class TableConfig:
         """
         df = df.with_columns(subject_id=self.subject_id_polars_expr)
         for name, node in self.cols.items():
-            df = df.with_columns(**{name: node.polars_expr})
+            df = df.with_columns(node.polars_expr.alias(name))
         return df
 
     def extract_events(
--- a/src/MEDS_extract/config.py
+++ b/src/MEDS_extract/config.py
@@ -590,11 +590,6 @@ class TableConfig:
             return expr.reinterpret(signed=True)
         return expr.cast(pl.Int64, strict=True)
 
-    @cached_property
-    def derived_column_exprs(self) -> dict[str, pl.Expr]:
-        """Polars expressions for each ``_table.cols`` derived column."""
-        return {k: v.polars_expr for k, v in self.cols.items()}
-
     @property
     def col_outputs(self) -> set[str]:
         """Column names produced by ``_table.cols`` — derived, not read from source."""
@@ -669,6 +664,12 @@ class TableConfig:
         stage-specific input directory for the join target and are applied by
         the caller via :meth:`JoinConfig.apply`.
 
+        Derived columns are applied in insertion order via a chain of
+        ``with_columns`` calls, so a later entry may reference a name defined
+        above it in the same ``_table.cols`` block. A forward reference (a
+        column referencing a name defined below it) fails at query time with
+        polars' standard missing-column error — intentionally not special-cased.
+
         Examples:
             >>> _ = pl.Config.set_tbl_width_chars(600)
             >>> tc = TableConfig.parse("t", {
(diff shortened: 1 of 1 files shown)


src/MEDS_extract/config.py
```diff
@@ -691,10 +692,42 @@ def prepare(self, df: pl.LazyFrame) -> pl.LazyFrame:
...
+            ╞════════════╪═══════════════╪══════════════╡
+            │ 100        ┆ 1990          ┆ 34           │
+            │ 200        ┆ 1996          ┆ 28           │
+            └────────────┴───────────────┴──────────────┘
         """
         df = df.with_columns(subject_id=self.subject_id_polars_expr)
-        if self.derived_column_exprs:
-            df = df.with_columns(**self.derived_column_exprs)
+        for name, node in self.cols.items():
+            df = df.with_columns(**{name: node.polars_expr})
```
mmcdermott: maybe just use `df.with_columns(node.polars_expr.alias(name))` instead of the single-dict dereferenece?

mmcdermott (commented):
Merge after this comment is addressed and CI passes, as otherwise it LGTM.

src/MEDS_extract/config.py
mmcdermott: Done in 8800bb2 — cleaner read, identical semantics. Thanks!

Issue #91, opened 2026-04-24
`HTTPSource`: support custom request headers (e.g. API-key auth)


mmcdermott:
## Motivation

AUMCdb is hosted on DANS Data Stations (a DataVerse instance) and authenticates
via a custom HTTP header — `X-Dataverse-key: <api-key>` — rather than HTTP Basic
auth. Current
[`[HTTPSource._make_client](https://github.com/mmcdermott/MEDS_extract/blob/dev/src/MEDS_extract/download/backends/http.py#L155-L223)`](https://github.com/mmcdermott/MEDS_extract/blob/dev/src/MEDS_extract/download/backends/http.py#L155-L223)
threads `auth: tuple[str, str]` into `httpx.BasicAuth` and provides no path for
arbitrary headers. Without a way to add them, `AUMCdb_MEDS` cannot migrate off
its hand-rolled `curl -H ... | zipfile.extractall` download.py
([[AUMCdb_MEDS/src/AUMCdb_MEDS/download.py:54](https://github.com/rvandewater/AUMCdb_MEDS/blob/main/src/AUMCdb_MEDS/download.py#L54)](https://github.com/rvandewater/AUMCdb_MEDS/blob/main/src/AUMCdb_MEDS/download.py#L54)).

This is also broadly useful for any future dataset that authenticates with bearer
tokens, custom API keys, or `Accept:` content negotiation (e.g. DataVerse's
native file-listing API returns JSON with the right `Accept:` header).

## Proposal

Add a `headers: dict[str, str] | None = None` kwarg to both
`HTTPSource.__init__` and `HTTPSource._make_client`. Thread it into the
`httpx.Client` default headers — applies to every request the client issues,
both `list_files`-style `GET`s and streaming `.part` downloads.

Sketch:

```python
def _make_client(
    cls,
    auth: tuple[str, str] | None = None,
    headers: dict[str, str] | None = None,  # NEW
    timeout: tuple[float, float] = (10.0, 60.0),
    max_attempts: int = 5,
    transport: httpx.BaseTransport | None = None,
) -> httpx.Client:
    ...
    client_kwargs: dict = {
        "auth": httpx.BasicAuth(*auth) if auth else None,
        "headers": headers,  # NEW — httpx handles None gracefully
        "timeout": httpx.Timeout(...),
        "follow_redirects": True,
    }
```

And expose it through the MESSY `sources:` schema:

```yaml
sources:
  dataset:
    - type: http
      headers:
[...]

PR #96, opened 2026-04-24
Fetcher: optional post-fetch archive unpack (#92)

### LINKED ISSUE #92: Fetcher: optional post-fetch archive unpack (zip, tar.gz)
## Motivation

Two of the public ETLs ship their raw data as archives that need to be expanded
before the rest of the pipeline can see the contained CSV/parquet files:

- **AUMCdb** ships the whole dataset as a single `.zip`. The current
  [`[download.py:61-63](https://github.com/rvandewater/AUMCdb_MEDS/blob/main/src/AUMCdb_MEDS/download.py#L61-L63)`](https://github.com/rvandewater/AUMCdb_MEDS/blob/main/src/AUMCdb_MEDS/download.py#L61-L63)
  calls `zipfile.ZipFile(...).extractall(output_dir)` after download.
- **HIRID** ships as `.tar.gz`. Extraction currently lives in
  [`[pre_MEDS.py:190-212](https://github.com/rvandewater/HIRID_MEDS/blob/main/src/HIRID_MEDS/pre_MEDS.py#L190-L212)`](https://github.com/rvandewater/HIRID_MEDS/blob/main/src/HIRID_MEDS/pre_MEDS.py#L190-L212),
  with a selective-member pattern (only `.csv` / `.parquet` members extracted).

In both cases, archive expansion is logically part of *"get raw files to disk"*
rather than *"transform raw files"* — it should happen at the fetcher layer.
Hoisting it there (a) centralizes the implementation across ETLs, (b) means
`resolve_source_files` / `get_supported_fp` / the shard stages see a flat
directory regardless of archive shape, and (c) removes archive handling from
`pre_MEDS.py` (HIRID) or the download script (AUMCdb), bringing us closer to
"all seven ETLs need zero custom download code".

## Proposal

Add an optional `unarchive` field on each source entry (or per-URL entry for
`HTTPSource`):

- `None` / absent — no unpack (default, unchanged)
- `"auto"` — infer from extension (`.zip` → zip, `.tar.gz`/`.tgz` → tar.gz,
  `.tar` → tar)
- `"zip" | "tar.gz" | "tar"` — explicit

Honor it in the base `Source.fetch` so every transport inherits the behavior
after its SHA-256 verify + atomic rename:

```python
# MEDS_extract/download/source.py  (sketch)
def fetch(self, remote, dest, do_overwrite=False):
    ...  # existing overwrite guard
    self._fetch(remote, dest)
    if remote.unarchive:
        _safe_extract(dest, fmt=remote.unarchive, target=dest.parent)
        if remote.cleanup_archive:
            dest.unlink()
```

`_safe_extract` must:

1. Validate every member path: reject absolute paths, reject `..` components
   (zip-slip / tar-slip guard).
2. For tar.gz, support an optional `include: [pattern]` or `members: [...]`
   filter — HIRID wants only `*.csv` and `*.parquet`.
3. Use `tarfile.data_filter` / Python 3.12+ filter semantics where available.

### Config schema evolution

Extend `RemoteFile` and the dict-form HTTP URL entry:

```yaml
# AUMCdb — DataVerse ships a single zip
sources:
  dataset:
    - type: http
      headers:  # depends on companion issue (custom headers)
        X-Dataverse-key: ${oc.env:AUMCDB_API_KEY}
      urls:
        - url: https://lifesciences.datastations.nl/api/access/datafile/:persistentId?persistentId=doi:10.17026/dans-22u-f8vd
          rel_path: AUMCdb.zip
          unarchive: zip
          cleanup_archive: true

# HIRID — PhysioNet ships tar.gz
sources:
  dataset:
    - type: physionet
      base_url: https://physionet.org/files/hirid/1.1.1/
      username: ${oc.env:DATASET_DOWNLOAD_USERNAME}
      password: ${oc.env:DATASET_DOWNLOAD_PASSWORD}
      # Per-file override for the archive members in SHA256SUMS.txt:
      unarchive_members:
        "raw_stage/*.tar.gz":
          unarchive: tar.gz
          include: ["*.csv", "*.parquet"]
```

The HIRID case is trickier because `PhysioNetSource.list_files` discovers file
entries from `SHA256SUMS.txt` rather than the YAML config, so the unarchive
metadata needs to be attached at enumeration time — either via a per-prefix
override map (sketched above) or by teaching `PhysioNetSource` to auto-apply
`unarchive: auto` when a discovered `rel_path` ends in a recognized archive
extension.

For the initial cut, recommend `unarchive: auto` at the `PhysioNetSource`
level (kwarg passed to `__init__`) rather than per-member; HIRID's selective
filter can land separately if anyone actually needs to skip members.

## Acceptance criteria

- [ ] `unarchive` field on `RemoteFile`, honored by `Source.fetch` after verify
- [ ] zip support (AUMCdb use case — full directory extraction)
- [ ] tar.gz support (HIRID use case)
- [ ] Path-traversal guard tested (zip-slip, absolute-path members rejected)
- [ ] `cleanup_archive` flag (default: true? discuss in PR)
- [ ] Dispatcher at `download/dispatch.py` forwards the field verbatim
- [ ] Doctest / unit test for each archive type + the traversal rejection
- [ ] `HTTPSource.list_files` honors per-entry `unarchive` on dict URL entries
- [ ] `PhysioNetSource` honors a top-level `unarchive="auto"` constructor kwarg

## Open questions

1. **Should unpack happen in-place or into a staged dir?** In-place means
   `dest.parent` grows arbitrarily; a staged `dest.parent / dest.stem` dir is
   cleaner but requires downstream path resolution to know about it. Recommend
   in-place for the first cut — matches AUMCdb's current behavior.
2. **Member filtering for tar.gz** — is the HIRID "only csv/parquet" filter
   necessary, or is extracting everything fine (just wastes disk)? If the
   extra members are genuinely junk, defer the filter to a follow-up.

## Related

- Custom HTTP headers (companion issue) — AUMCdb needs both to migrate. (#91)
- `Medical-Event-Data-Standard/MEDS_extract` issue #81 (MESSY `sources:` block).

## Alternative
Alternatively, we can also just support these archived file types as readable inputs, that way the system can navigate the archive versions fine during downstream processing (this is how csv.gz compressed files are handled)

mmcdermott:
## Summary

- New `MEDS_extract.download.unarchive` module providing `safe_extract` with zip-slip / tar-slip / symlink-escape guards.
- `RemoteFile` gets `unarchive` + `cleanup_archive` fields; `Source.fetch` calls `safe_extract` after `_fetch` when set.
- Supported formats: `zip`, `tar`, `tar.gz` / `tgz`, and `auto` (infer from suffix; no-op on non-archive extensions).
- `HTTPSource` dict URL entries forward `unarchive` / `cleanup_archive` verbatim; `PhysioNetSource` gains a top-level `unarchive="auto"` kwarg that stamps every listed `RemoteFile` (HIRID-shaped releases).
- Motivating cases: AUMCdb ships its entire dataset as a DataVerse zip; HIRID ships raw data as `.tar.gz` members inside the PhysioNet release. With this, both migrate off bespoke `extractall` paths in their ETL's pre-MEDS step.

Closes #92. **Stacks on #91** — set to `base: feat/http-source-headers`. Retarget to `dev` after #91 merges.

## Safety posture

Every archive member's path is validated BEFORE any bytes land on disk. Absolute paths, `..` escapes, and symlink/hardlink targets pointing outside the target are rejected with a clear `ValueError` naming the offending member. Tar extraction also uses PEP 706's `data_filter` as defense-in-depth — strips setuid/setgid/sticky, rejects special files (devices, fifos).

## Scope notes

- **In-place extraction** into `dest.parent` (matches AUMCdb's current hand-rolled `extractall`). A staging dir would be cleaner but requires downstream path resolution to know about it — punted to follow-up.
- **Member filtering** (HIRID's "only csv/parquet") deferred per open question #2 on the issue. Extracting everything wastes disk but is otherwise harmless.
- **`cleanup_archive` default is `False`** so re-runs skip redownload via the manifest cache. Flip to `True` for large archives (AUMCdb full release).
- **`do_overwrite` only clears the archive + `.part`**, not previously-extracted members; documented as a caller responsibility.
- **Dataset output collisions** are still caught by `Fetcher._resolve_dest` since extraction lands on disk after `_fetch` returns — dest paths are whatever the manifest declares.

## Test plan

- [x] Doctest for each archive type (zip, tar.gz) on `safe_extract` — full round-trip.
- [x] Doctest rejecting zip-slip (`../escaped.txt`), absolute-path (`/etc/passwd-hijack`), and tar-slip members. Verifies target directory stays empty after rejection.
- [x] Doctest for `infer_format` / `resolve_format` — archive extensions and no-op cases (`patients.csv.gz`, `README.md`).
- [x] End-to-end doctest on `Source.fetch` using a stub that synthesizes a zip, extracts it, and removes the archive via `cleanup_archive=True`.
- [x] `HTTPSource` class-level doctest locks in per-entry `unarchive` / `cleanup_archive` passthrough.
- [x] `PhysioNetSource` class-level doctest locks in the `unarchive="auto"` kwarg.
- [x] `pytest --doctest-modules -m "not integration"` — 123 passed (up from 118).

## Follow-ups (separate issues / PRs)

- Member filter for `tar.gz` (HIRID's `*.csv` / `*.parquet` pattern) if/when someone actually needs it — issue #92 open-question 2.
- `_already_complete` interaction: if `cleanup_archive=True` removed the archive on the previous run, the next `fetch_all` pass will re-download it. Could leave a marker file or extend the "already done" check to recognize extracted members; deferred.
- `do_overwrite` cleaning previously-extracted members; currently a caller responsibility.


mmcdermott:
Integration test failures here are the PhysioNet rate-limit flake (`httpx.ConnectTimeout` after partial successes). Fixed in #100 (lowers `concurrency=8` → `concurrency=4`, which is what the `Fetcher` docstring already calls the polite floor). Once #100 lands on dev, a rebase/merge here will pick up the fix and the integration job should turn green.

mmcdermott:
## Re: archive-extraction library research

Question raised in conversation: "are there not existing libraries we can import and rely on to do some of this? Why do we need our own code?"

**Short answer**: we DO rely on existing libraries — the ones we use are stdlib `zipfile` and `tarfile`. The ~50 lines in `MEDS_extract/download/unarchive.py` are a *path-validation wrapper* around their `extractall()`, not a re-implementation of zip/tar parsing.

### What we use vs. what we wrote

| Concern | Library | New code? |
|---|---|---|
| Parse + extract `.zip` | stdlib `zipfile.ZipFile` | no — direct call |
| Parse + extract `.tar.gz` / `.tar` / `.tgz` | stdlib `tarfile.open(..., "r:gz" / "r:")` | no — direct call |
| Reject special tar members (devices, fifos), strip setuid/setgid | stdlib `tarfile`'s `data` filter (PEP 706 — official CVE-2007-4559 fix) | no — `members=..., filter="data"` |
| Pre-validate every member's path against zip-slip / tar-slip / symlink escape | — (see "why" below) | **yes**, ~30 LOC |
| Format dispatch (`auto` → infer from suffix) | — | yes, ~15 LOC |

### Third-party libraries we considered and rejected

- **`libarchive`** (via [`python-libarchive`](https://pypi.org/project/python-libarchive/) or `libarchive-c`): native C performance and ~30 supported formats. **Rejected**: needs the system `libarchive` shared lib installed (extra OS-level install friction for users on minimal containers / shared HPC), and our entire format requirement is two formats. The wins (perf, breadth) don't matter for a one-shot extraction of CSVs that came over the wire — extraction time is dominated by disk I/O and is already <1s for the worst-case archives we'll see.
- **[`patool`](https://github.com/wummel/patool)** / [`pyunpack`](https://github.com/ponty/pyunpack): shell out to `7z`, `unrar`, etc. **Rejected**: needs external binaries on `PATH`, runs subprocesses (extra failure mode + harder-to-diagnose errors), and again only buys format breadth we don't need.
- **[`extractcode`](https://pypi.org/project/extractcode/)**: kitchen-sink wrapper over libarchive + 7z + stdlib. **Rejected**: same dep-weight issues amplified, and pulls in license-scanning code we don't want.

### Why the path-validation wrapper exists at all

`tarfile.data_filter` (Python 3.12+) IS robust against zip-slip / tar-slip — but it raises a generic `tarfile.FilterError` whose message doesn't tell an ETL author *which* member of *which* archive failed. Our pre-pass walks the member list, identifies the offending name (e.g. `'../escaped.txt'`), and raises a `ValueError` like `Refusing to extract unsafe zip member '../escaped.txt' from /tmp/X.zip: member path would land outside the extraction directory.` — which goes straight to a useful bug report. We keep `data_filter` as defense-in-depth in case our pre-pass ever regresses.

For zip, stdlib `zipfile` has *no* equivalent built-in safety filter, so the pre-pass is doing real work, not just diagnostics.

The same rationale is now baked into the module docstring at `src/MEDS_extract/download/unarchive.py` (commit b08763d) so it's visible to anyone reading the code.

src/MEDS_extract/download/source.py
```diff
@@ -109,6 +127,8 @@ class RemoteFile:
     rel_path: str
     size: int | None = None
     sha256: str | None = None
+    unarchive: str | None = None
```
mmcdermott, line 160: I think there should be a one-arg way to toggle on an auto mode that recognizes archive type by file suffix, unarchives it, and cleans up the archive. The way to do this is to have cleanup_archive also default to None, and unarchive can be a StrEnum that recognizes file suffixes or AUTO or None, and when it is auto if cleanup_archive is None (indicating unset) it does cleanup, but it can be set to True or False. What do you think?

src/MEDS_extract/download/unarchive.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+    if fmt not in {"zip", "tar", "tar.gz"}:
+        raise ValueError(f"safe_extract: unsupported fmt {fmt!r}; expected one of 'tar', 'tar.gz', 'zip'.")
+    if not archive_path.exists():
+        raise FileNotFoundError(f"safe_extract: archive does not exist: {archive_path}")
+    target_dir.mkdir(parents=True, exist_ok=True)
+    target_root = target_dir.resolve()
+
+    if fmt == "zip":
+        with zipfile.ZipFile(archive_path) as zf:
```
mmcdermott: I think from a design perspective, each format should probably be a different function or something, then this function can just use a switch case over a strenum to route to the right function. That'd make it a bit cleaner, right? Push back if you disagree.

src/MEDS_extract/download/unarchive.py
```diff
@@ -0,0 +1,339 @@
...
+        members = list(_iter_tar_members(tf))
+        for m in members:
+            if not _is_safe_member(m.name, target_root):
+                raise ValueError(
+                    f"Refusing to extract unsafe tar member {m.name!r} from {archive_path}: "
+                    "member path would land outside the extraction directory."
+                )
+            # A symlink's link target is a second escape vector: a safe member name
+            # ``good.txt`` pointing at ``../../etc/passwd`` would bypass the name check.
+            if (m.issym() or m.islnk()) and not _is_safe_member(m.linkname, target_root):
```
mmcdermott: why not just have this be part of `_is_safe`? 

Really, actually, the code for zipfile and tarfile seem nearly identical. If you had a generic base class for compressed file or something, it could have an overwriteable context manager and an _is_safe_member function that checks the name and, if the input is the right type, checks the symlink stuff, and then the actual extraction code would be shared. All you'd need to do between zip and tar would be change the context manager. Would that be cleaner?

mmcdermott: Done in 5c4c38f, but as a shared validator function rather than a base class. _validate_member handles both the name check AND the optional link-target check (the second-vector concern), so each per-format extractor just lists its members and calls validate for each. _extract_tar and _extract_tar_gz share their body via _extract_tar_with_mode. I went with module-level helpers over a base class because the only thing that varies between zip and tar is the context manager + the member adapter — a class would make that explicit but would also require subclasses to wire up an iterator that yields (name, link_target_or_None) tuples, which is more ceremony than the dispatch table for what's still a 3-format universe. Push back if you'd rather see the class shape.

src/MEDS_extract/download/source.py
mmcdermott: Done in 5c4c38f. cleanup_archive is now bool | None (default None); Source.fetch cleans up when unarchive=AUTO AND cleanup_archive is None, otherwise honors the explicit boolean. Both backends propagate the tri-state. unarchive is now an ArchiveFormat(StrEnum).

src/MEDS_extract/download/unarchive.py
mmcdermott: Done in 5c4c38f. Each format is its own _extract_zip / _extract_tar / _extract_tar_gz function; safe_extract dispatches via an _EXTRACTORS: {ArchiveFormat: callable} table. Adding a new format is now: add enum value + extractor function + table entry.

src/MEDS_extract/download/backends/physionet.py
```diff
@@ -61,6 +72,17 @@ class PhysioNetSource(HTTPSource):
         Traceback (most recent call last):
             ...
         ValueError: PhysioNetSource: username and password must be supplied together ...
+
+        ``unarchive`` / ``cleanup_archive`` propagate to every
+        :class:`~MEDS_extract.download.source.RemoteFile` listed. ``"auto"`` is the
+        expected value for HIRID-shaped releases where only some members are archives:
```
mmcdermott: Why call out HIRID specifically, here? This comment makes it seem like something here is designed for one specific example.

src/MEDS_extract/download/source.py
```diff
@@ -156,15 +184,58 @@ def list_files(self) -> Iterable[RemoteFile]:
...
+            ...         with _zipfile.ZipFile(buf, "w") as zf:
+            ...             for name, data in self._payload:
+            ...                 zf.writestr(name, data)
+            ...         dest.write_bytes(buf.getvalue())
+            >>>
+            >>> src = ZipSource([("a.csv", b"col\\n1"), ("sub/b.csv", b"col\\n2")])
+            >>> [remote] = list(src.list_files())
+            >>> with tempfile.TemporaryDirectory() as d:
+            ...     d = Path(d)
+            ...     src.fetch(remote, d / "bundle.zip")
```
mmcdermott: I think that here, you should use pretty-print-directory to make a more readable, clear doctest: https://github.com/mmcdermott/pretty-print-directory

src/MEDS_extract/download/source.py
```diff
@@ -156,15 +184,58 @@ def list_files(self) -> Iterable[RemoteFile]:
...
+            ...         return [RemoteFile("bundle.zip", unarchive="auto", cleanup_archive=True)]
+            ...     def _fetch(self, remote, dest):
+            ...         buf = io.BytesIO()
+            ...         with _zipfile.ZipFile(buf, "w") as zf:
+            ...             for name, data in self._payload:
+            ...                 zf.writestr(name, data)
+            ...         dest.write_bytes(buf.getvalue())
+            >>>
+            >>> src = ZipSource([("a.csv", b"col\\n1"), ("sub/b.csv", b"col\\n2")])
+            >>> [remote] = list(src.list_files())
```
mmcdermott: This makes me a bit confused re API -- can you respond to this comment and clarify design? Here, you are getting a file from the src, then later doing src.fetch with that remote file as an input, and with an output path as an input -- why do we need to do `fetch` on the `src` when we're passing the remote file gotten from the `src`? Why not just `remote.fetch()`? Also, given the archiving here is a directory, it seems like the final output path is really being ignored here -- is that an intuitive API?

src/MEDS_extract/download/source.py
```diff
@@ -178,6 +249,25 @@ def fetch(self, remote: RemoteFile, dest: Path, do_overwrite: bool = False) -> N
                 f"Pass do_overwrite=True to force a refetch, or delete the file first."
             )
         self._fetch(remote, dest)
+        if remote.unarchive:
+            # Import locally so the ABC stays importable without the optional extras —
+            # ``unarchive.py`` uses only stdlib, but keeping the import lazy matches the
+            # pattern the rest of this module uses for transport-specific deps.
+            from .unarchive import ArchiveFormat, resolve_format, safe_extract
```
mmcdermott: I would say as a general rule, if there aren't specific dependencies needed just do the import at the top of the file.

src/MEDS_extract/download/unarchive.py
```diff
@@ -0,0 +1,438 @@
...
+        >>> ArchiveFormat("rar")
+        Traceback (most recent call last):
+            ...
+        ValueError: 'rar' is not a valid ArchiveFormat
+    """
+
+    AUTO = "auto"
+    ZIP = "zip"
+    TAR = "tar"
+    TAR_GZ = "tar.gz"
```
mmcdermott, line 96: Do we need both this and tgz, even though we assign tgz to archive format .TAR_GZ?

src/MEDS_extract/download/unarchive.py
```diff
@@ -0,0 +1,438 @@
...
+        return False
+    resolved = (target_root / Path(*pure.parts)).resolve()
+    try:
+        resolved.relative_to(target_root)
+    except ValueError:
+        return False
+    return True
+
+
+def _extract_zip(archive_path: Path, target_root: Path) -> None:
```
mmcdermott, line 263: I thought you were going to try to unify these all into one function to remove some repeated code?

src/MEDS_extract/download/source.py
mmcdermott: Will do — `print_directory` is much more readable than the `rglob + sorted` walk. I'm going to move this doctest as part of addressing the API question two threads down (probably to `Fetcher` rather than `Source.fetch`), and switch to `print_directory` while I'm there.

src/MEDS_extract/download/unarchive.py
mmcdermott: Fair — the `_extract_tar_with_mode` shim only collapses tar vs tar.gz, not zip vs tar. The blocker for full unification is that zipfile and tarfile present different APIs (different context manager, different member iterators with different metadata shapes — TarInfo carries linkname, ZipInfo doesn't, etc.).

The clean way to unify is a small `_ArchiveAdapter` shape:

```python
@dataclass
class _ArchiveAdapter:
    open: Callable[[Path], ContextManager[Any]]   # returns ZipFile / TarFile
    iter_members: Callable[[Any], Iterable[_Member]]  # _Member = (name, link_target_or_None, is_special)
    extract: Callable[[Any, Path, list[_Member]], None]   # delegates to extractall + filter
```

Then `safe_extract` is one function that takes an adapter, walks members, validates, and extracts. `_ZIP_ADAPTER` and `_TAR_ADAPTER` are the two concrete instances (TAR_GZ shares the tar adapter with a different open mode — a partial). Will refactor into that shape in the next push.

src/MEDS_extract/download/source.py
mmcdermott: Right — the local import was a copy of the `from __future__ import annotations` pattern from elsewhere in the module, but `unarchive.py` is stdlib-only and there's no extras-related reason to keep it lazy. It also goes away entirely if we move unarchive to `Fetcher` (per the thread above). Will fix in the next push either way.

src/MEDS_extract/download/source.py
mmcdermott: Real architectural question — let me lay out my thinking and propose a fix:

**Why `src.fetch(remote, dest)` not `remote.fetch(dest)`?**

`RemoteFile` is a frozen dataclass on purpose — pure transport-agnostic data (a manifest entry: rel_path, size, sha256, optional unarchive metadata). It can be serialized, passed across process boundaries, compared for equality, etc. The fetch logic depends on the transport (httpx for HTTP, fsspec for S3/local, future DataverseSource via the DataVerse API) and lives on `Source` for that reason.

To support `remote.fetch(dest)` we'd need to attach a back-reference to the originating `Source` on every `RemoteFile`. That couples the data to a live source instance and breaks the serializable-data property — and in practice the user always has both a `Source` and the `RemoteFile`s it produced, since they got the latter from the former. So I'd defend `src.fetch(remote, dest)`. Let me know if you'd still rather flip it.

**Why does `dest` look ignored when an archive is involved?**

Because it kind of is, today, and that's the real bug in the API. When `unarchive` is set, `dest` is just a transient staging location for the archive bytes; the canonical outputs are extracted members under `dest.parent`, and with `cleanup_archive=True` the staging file doesn't exist after the call.

The right fix is to **move unarchive out of `Source.fetch` and into `Fetcher`**. Then:

- `Source.fetch(remote, dest)` keeps clean semantics: this exact file lands at this exact path, every time, no exceptions.
- `Fetcher._fetch_one` calls `Source.fetch`, then conditionally calls `safe_extract(dest, dest.parent, ...)` and (if applicable) cleans up the archive. The "the dest path becomes a directory's worth of stuff" semantic now lives in the *orchestrator* layer where it belongs, with explicit documentation.
- Users who hand-roll a `Source` consumer (no `Fetcher`) get pure file-fetch semantics without the archive surprise.
- The unarchive/cleanup metadata still travels on `RemoteFile` (it's manifest-level info, and Fetcher needs to read it).

I think this is the right move. About to push it unless you push back. The Fetcher doctest is where the `print_directory`-based example will live.

src/MEDS_extract/download/unarchive.py
mmcdermott: No, you're right — having both is redundant. The reason I added `TGZ` was so users writing `unarchive: tgz` in YAML round-trip cleanly through the enum. But that's better handled by normalizing `'tgz' → 'tar.gz'` *before* constructing the enum, so the enum stays as { AUTO, ZIP, TAR, TAR_GZ } — four canonical values, no aliases. Will drop `ArchiveFormat.TGZ` and add the normalization in `resolve_format`.

src/MEDS_extract/download/backends/physionet.py
mmcdermott: Fair point — fixed in the next push. Rephrased as 'releases that ship a mix of archive and non-archive members' so the docstring describes the general pattern rather than implying the feature was carved out for one dataset.

mmcdermott:
## Design proposal: drop per-file `Source.fetch` + `RemoteFile` from public API; keep Fetcher

Walking through this thread we cycled through five design alternatives. Posting the summary here so the rationale is visible alongside the code; the implementation is up at **#101** (targeting this branch as base, so the diff is reviewable as a delta on top of the unarchive feature).

### The two design smells

1. **`Source.fetch(remote, dest)` had ambiguous semantics under unarchive.** When `remote.unarchive` was set, the canonical outputs were extracted members under `dest.parent`; `dest` was transient and sometimes deleted. The "exact file at exact path" contract was a half-truth.
2. **Per-file granularity wasn't used by anyone.** Every consumer (CLI, tests, planned `DataverseSource`) downloads the whole bundle from a Source, then immediately walks away. The `Source` ↔ `Fetcher` decoupling looked like separation-of-concerns but was actually two-class boilerplate buying nothing.

### Alternatives we considered

| # | Shape | Verdict |
|---|---|---|
| A | Status quo: `src.fetch(remote, dest)` + document `dest` weirdness for archives | Rejected: doesn't fix the smell, leaves the `(src_b, remote_from_src_a)` mispair footgun |
| B | `src.fetch(name, dest)` — replace `RemoteFile` with a plain string | Rejected: forces `Source` to grow a `stat(name)` API for the skip-check, churn for no real win |
| C | `remote.fetch(dest)` — `RemoteFile` carries a back-ref to its source | Rejected (closer): clean ergonomics, but `RemoteFile` carries a Source ref it doesn't conceptually need; users still see `RemoteFile` as a "thing they handle" |
| D | Drop per-file fetch entirely, also drop `Fetcher`, fold orchestration into each backend | Rejected: lose the shared SIGINT-cancel / skip-check / unarchive policy code, every backend reimplements it |
| E (winning) | Drop per-file public fetch, drop `RemoteFile` from public surface; **inject `Fetcher` into `Source` at construction; `Source.download_all(dest)` is the only public fetch method** | Adopted |

### Why E wins

- **`Fetcher` keeps doing what it actually does today** — concurrency policy, skip-check, sha verify, post-fetch unarchive — but stops needing a `dest_dir` at construction (the dest comes from `download_all`'s arg) and stops being a public orchestrator that users call directly.
- **`Source.download_all(dest_dir)` is the single public fetch method.** No per-file API to confuse with archives. The "dest gets replaced by a directory's worth of stuff" is now just documented behavior on `download_all`, where it belongs (orchestrator scope, not transport scope).
- **`RemoteFile` becomes truly internal.** It's still the `_list_files()` return type (Fetcher needs it for size/sha/unarchive metadata), but it's dropped from the package `__init__` exports. Stub sources in tests construct it; regular users never see one.
- **One `Fetcher` for many `Source`s.** The shared-policy use case (a single CLI invocation with N sources) works naturally — build one `Fetcher`, hand it to every source via `sources_from_spec(spec, fetcher=fetcher)`. CLI becomes:

  ```python
  fetcher = Fetcher(max_concurrency=cfg.concurrency, ...)
  sources = sources_from_spec(spec_dict, key=cfg.key, fetcher=fetcher)
  for src in sources:
      src.download_all(raw_input_dir)
  ```

- **`(src_b, remote_from_src_a)` footgun gone.** No user code path can produce a wrong-pair anymore — there's no `(source, remote)` tuple to mis-pair.
- **Backends barely change.** Every `_list_files` + `_fetch` body stays identical. The renames are: `list_files` → `_list_files` (now private); `Source.__init__` accepts `fetcher=`; the public `Source.fetch` deletes.

### What this preserves from the existing PR

Every robustness invariant from #88 / #92:
- SIGINT-cancel-futures with `wait=False` (the painstakingly-debugged thing).
- Atomic `.part` rename, range-resume, `Content-Range` validation.
- Manifest-collision rejection before thread pool spin-up.
- Skip-on-existing via size + sha cheap-then-expensive check.
- Unarchive (now from `Fetcher._maybe_unarchive`, called per-file in the worker thread) — same `safe_extract`, same tri-state `cleanup_archive`, same doctests.

### What this PR (#101) does NOT do

The four other open review items on this branch — HIRID-specific phrasing in the docstring (✓ already addressed), drop `TGZ` enum value, unify zip/tar via adapter, top-level imports — are still valid and still on this branch. After #101 lands, I'll fold those in here too. They're orthogonal to the architectural change.

### Merge order proposal

1. Review #101 against this branch.
2. Merge #101 into `feat/fetcher-unarchive` (this branch).
3. Address the four remaining review threads on this branch.
4. Retarget this branch to `dev` and merge.

LMK if a different sequencing reads better.

mmcdermott:
Superseded by #103 (download API redesign) + #104 (archive feature on top). The original feat/fetcher-unarchive branch had become entangled — these split it into a clean reviewable stack: review #103 first, then #104 as a delta of just the archive functionality.

PR #97, opened 2026-04-24
Fix #51: poll for complete parquets in extract_code_metadata reducer

### LINKED ISSUE #51: Stage extract_code_metadata occasionnally fails when N_WORKERS >1 with a Polars Exception error
When running with multiple workers (N_WORKERS > 1), the pipeline occasionally fails during `extract_code_metadata` with a Polars parquet read error:

```
polars.exceptions.ComputeError: parquet: File out of specification:
A parquet file must contain a header and footer with at least 12 bytes
```
This looks like we sometimes try to read a partially-written/corrupted parquet. I haven’t been able to reproduce it deterministically, but it appears intermittently.

It might come from `extract_code_metadata.py` line 442 :

```
    while not all(fp.exists() for fp in all_out_fps):  # pragma: no cover
        missing_files_str = "\n".join(f"  - {fp.resolve()!s}" for fp in all_out_fps if not fp.exists())
        logger.info(f"Waiting to begin reduction for all files to be written...\n{missing_files_str}")
        time.sleep(cfg.polling_time)
```
and that `fp.exists()` could be replaced with a stronger criterion such as `is_complete_parquet_file(fp)` from `MEDS_transforms/mapreduce/rwlock.py `

I've faced it when running the updated [MIMIC_IV_MEDS](https://github.com/Medical-Event-Data-Standard/MIMIC_IV_MEDS) pipeline with latest version of MEDS_extract, see [PR](https://github.com/Medical-Event-Data-Standard/MIMIC_IV_MEDS/pull/32).

mmcdermott:
Closes #51.

## Summary

Two layered fixes for the `N_WORKERS>1` Polars race:

1. **Atomic temp+rename writes** (`atomic_write_parquet`): replaces upstream `MEDS_transforms.dataframe.write_df` (which calls `write_parquet` directly with `O_TRUNC`) at the stage's three `rwlock_wrap` call sites. The destination path now atomically appears as the fully-written parquet — no zero-byte `O_TRUNC` window for a concurrent reader to observe.
2. **Stronger polling check** (`wait_for_complete_parquets`): replaces `Path.exists()` with `is_complete_parquet_file` — the same parquet-completeness probe `rwlock_wrap`'s own `default_file_checker` uses. Stays as defense-in-depth for any non-stage-managed parquet that might land in `partial_metadata_dir` via a different code path.

Either fix alone closes the symptom from #51:

```
polars.exceptions.ComputeError: parquet: File out of specification:
A parquet file must contain a header and footer with at least 12 bytes
```

…but the atomic write is the cleaner root-cause fix. The polling fix is kept because it's cheap and matches what the rest of the module already does for "is this output done?" decisions.

## Why this races (root cause)

`pl.DataFrame.write_parquet` opens its destination with `O_TRUNC` and writes in-place — no temp file, no atomic rename. Verified with `strace` (single `openat(..., O_WRONLY|O_CREAT|O_TRUNC)`, no `rename()`) and by polling sizes during a 2 MB write: `{0, 324_561, 817_889, 1_310_720, ..., 2_112_404}`. The destination path is observable as a zero-byte (or partial) file for a measurable window.

The stage's `rwlock_wrap` calls already prevent two writers from clobbering each other (file-level `FileLock`), but the reducer's reads weren't guarded — and the polling exit check used the weakest possible criterion (`Path.exists()`).

## Test plan

- [x] `test_reducer_does_not_crash_on_partial_parquet_from_concurrent_worker` — exercises the production polling helper directly. Verified fails against a reverted `Path.exists()` polling and passes against the fix.
- [x] `test_atomic_write_parquet_never_exposes_partial_state` — spawns a writer thread that runs `atomic_write_parquet` on a 10k-row frame and a watcher thread that samples the destination path as fast as Python allows. Verified fails (16/220 partial observations) when the production helper is reverted to the non-atomic `write_df` implementation; passes (220/220) with atomic-write.
- [x] All existing `extract_code_metadata` in-process pipeline tests still pass — no behavior change for users.
- [x] Full test suite (`pytest --doctest-modules -m "not integration"`) — 120 passed.

## Why a sequence-reproduction test, not a multi-process one

The race requires two OS-level workers talking through a shared filesystem. There's no deterministic way to schedule subprocess B's `write_parquet` to be mid-flush at the exact instant subprocess A's polling loop probes. The thread-based fixtures model the exact observable state (`O_TRUNC` window for the polling test; concurrent stat-while-writing for the atomic-write test) without the timing noise of subprocesses.

I extracted the polling loop into a `wait_for_complete_parquets` helper specifically so the test exercises the actual production code, not a copy. Without that extraction a future regression that re-introduces `Path.exists()` would only manifest in the (intermittent, multi-worker) integration test.

## Note on integration test failure

The `integration (3.12)` failure on this PR is the orthogonal PhysioNet rate-limit flake at `concurrency=8` — fixed in #100, which lowers the test's concurrency to 4. After #100 lands, a rebase here will pick it up and the integration job will turn green.

### DIFF
--- a/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
+++ b/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
@@ -348,7 +348,7 @@ def wait_for_complete_parquets(fps: list[Path], polling_time: float) -> None:
     caller is responsible for any timeout (in practice a Hydra job timeout
     or a user ``Ctrl-C``).
     """
-    while not all(is_complete_parquet_file(fp) for fp in fps):  # pragma: no cover
+    while not all(is_complete_parquet_file(fp) for fp in fps):
         missing_files_str = "\n".join(
             f"  - {fp.resolve()!s}" for fp in fps if not is_complete_parquet_file(fp)
         )
--- a/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
+++ b/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
@@ -12,7 +12,7 @@ import polars as pl
 from dftly import Parser
 from meds import CodeMetadataSchema
 from MEDS_transforms.dataframe import write_df
-from MEDS_transforms.mapreduce.rwlock import rwlock_wrap
+from MEDS_transforms.mapreduce.rwlock import is_complete_parquet_file, rwlock_wrap
 from MEDS_transforms.parser import cfg_to_expr
 from MEDS_transforms.stages import Stage
 from omegaconf import DictConfig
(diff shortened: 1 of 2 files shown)


mmcdermott:
Integration test failures here are the PhysioNet rate-limit flake (`httpx.ConnectTimeout` after partial successes). Fixed in #100 (lowers `concurrency=8` → `concurrency=4`, which is what the `Fetcher` docstring already calls the polite floor). Once #100 lands on dev, a rebase/merge here will pick up the fix and the integration job should turn green.

mmcdermott:
Re: "can we import the atomic write fn from MEDS-transforms directly?"

Checked — MEDS-transforms doesn't ship one today. `MEDS_transforms.dataframe.write_fn.write_df` is the only writer it exposes, and it calls `pl.DataFrame.write_parquet` directly with `O_TRUNC` semantics (i.e. it's the *non*-atomic version this PR works around).

Filed [mmcdermott/MEDS_transforms#392](https://github.com/mmcdermott/MEDS_transforms/issues/392) proposing they make `write_df` itself atomic via tmp+rename — that way every consumer (not just MEDS_extract) gets the safety guarantee for free, and we can drop our local `atomic_write_parquet` helper here in a follow-up once it lands. Until then this PR keeps the local helper.

Will also watch for the PhysioNet flake on this PR's integration job and harden it (further concurrency drop / retries) if it returns now that #100 is merged in via the dev rebase.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -296,6 +295,67 @@ def extract_all_metadata(
...
+    ``tests/test_extract_code_metadata_race.py`` can call it directly — that
+    way the test exercises the actual production polling logic, not a copy
+    that could drift if the polling check is changed in one place but not
+    the other.
+
+    Sleeps :data:`polling_time` seconds between checks. Loops forever; the
+    caller is responsible for any timeout (in practice a Hydra job timeout
+    or a user ``Ctrl-C``).
+    """
+    while not all(is_complete_parquet_file(fp) for fp in fps):  # pragma: no cover
```
mmcdermott: you shouldn't need a nocvoer here.

mmcdermott: meaning that your regression test should cover exactly these lines, right?

mmcdermott: Right — the regression test exercises this loop body directly. Removed in d300720.

PR #101, opened 2026-04-27
Redesign #96 download API: Source.download_all + pool-as-arg + DownloadPolicy

mmcdermott:
Stacks on PR #96 (base = `feat/fetcher-unarchive`). Architectural redesign of the download layer's public surface, converged on through the [#96 review threads](https://github.com/mmcdermott/MEDS_extract/pull/96#issuecomment-4328412350).

> **Note**: this PR's first version had a `Fetcher` class. After further review (see comment thread on this PR), `Fetcher` was identified as a parameter object pretending to be a class and was deleted. The current shape is **pool-as-argument + `DownloadPolicy`** — see "API delta" below.

## Why the redesign

Two design smells in #96 that compound:

1. **Per-file fetch had ambiguous semantics under unarchive**: `Source.fetch(remote, dest)` claimed "this exact file lands at this exact path", but when `remote.unarchive` was set the canonical outputs were extracted members under `dest.parent` — `dest` was transient, sometimes deleted. The dest path was effectively ignored.
2. **Per-file granularity wasn't used by anyone**: every consumer (CLI, tests, planned `DataverseSource`) downloads the whole bundle. The `Source` ↔ `Fetcher` decoupling was theatre — every code path paired them 1:1.

We cycled through five alternatives in the #96 review thread before landing here: drop public per-file fetch, drop public `RemoteFile`, drop `Fetcher` class, **make `Source.download_all(dest, *, pool=None, policy=None)` the only public fetch entry point**.

## API delta

| Before (#96 base) | After (this PR) |
|---|---|
| `Fetcher(dest, max_concurrency=4, ...)` | gone — `Fetcher` class deleted |
| `fetcher.fetch_all(source) -> FetchReport` | gone |
| `Source.fetch(remote, dest, do_overwrite=...)` | gone — no public per-file fetch |
| n/a | **`DownloadPolicy(continue_on_error=False, do_overwrite=False)`** — frozen config bag |
| n/a | **`Source.download_all(dest, *, pool=None, policy=None) -> FetchReport`** — single public fetch method |
| `source.list_files()` | `source._list_files()` — private hook |
| `RemoteFile` exported from `MEDS_extract.download` | dropped from `__init__`; lives in `source.py`, importable for tests |
| `sources_from_spec(spec, key=...)` | unchanged (no `fetcher=` plumbing) |

## Pool ownership semantics

`Source.download_all` borrows pools, never owns them — except when no pool is passed:

- **`src.download_all(dest)`** — `download_all` builds a private 4-worker pool, runs the bundle, tears down via `shutdown(wait=False, cancel_futures=True)` (the SIGINT-safe semantics from PR #88, preserved).
- **`src.download_all(dest, pool=mine)`** — caller owns the pool. Useful when one CLI invocation drives multiple sources; the CLI builds one pool via `ExitStack`, all sources share it.

This is the actual shared-pool benefit the previous `Fetcher` class never realized — `Fetcher.fetch_all` constructed a fresh `ThreadPoolExecutor` on every call, so even with one shared `Fetcher` instance, every source got its own pool. With pool-as-arg, threads from a finished source's manifest become available to the next source under the same pool.

## CLI

```python
policy = DownloadPolicy(continue_on_error=cfg.continue_on_error, do_overwrite=cfg.do_overwrite)
with ExitStack() as stack:
    pool = stack.enter_context(ThreadPoolExecutor(max_workers=cfg.concurrency))
    for source in sources:
        stack.enter_context(source)
    for source in sources:
        report = source.download_all(raw_input_dir, pool=pool, policy=policy)
        ...
```

## What's preserved from #96

Every robustness invariant from PR #88 / #92:

- SIGINT-safe `cancel_futures` shutdown — on the private-pool path. Caller-owned pools are caller responsibility.
- Atomic `.part` rename, range-resume, `Content-Range` validation.
- Manifest-collision rejection before pool submission.
- Skip-on-existing via size + sha cheap-then-expensive check.
- Unarchive: `safe_extract` is called from `_maybe_unarchive` (per-file in the worker thread). Same tri-state `cleanup_archive`, same doctests.

## Module structure

One file (`source.py`, ~580 lines) holds the manifest types, policy/report dataclasses, orchestration helpers (free functions), and the `Source` ABC. The previous `_types.py` + `fetcher.py` split was needed to break a cycle that existed because `Source.__init__` constructed a `Fetcher()` as default. With `Fetcher` gone, no cycle, no third file.

`unarchive.py` (~440 lines) is unchanged — independent of this redesign.

## Test plan

- [x] All 126 non-integration tests pass (was 126 on #96 base).
- [x] `pre-commit` clean across every changed file.
- [x] `Source.download_all` end-to-end doctest exercises both the private-pool and caller-owned-pool paths.
- [x] CLI test (`test_meds_extract_download_cli_end_to_end`) — passes with the new shared-pool path.
- [x] SIGINT regression test (`tests/_fetcher_sigint_child.py`) — passes; uses the private-pool path so the `cancel_futures` invariant is the one under test.
- [ ] Integration test (`test_example.py`) — depends on PhysioNet, will run in CI.

## Diff summary vs PR base (`feat/fetcher-unarchive`)

```
9 files changed, 573 insertions(+), 684 deletions(-)
src/MEDS_extract/download/fetcher.py    | 413 deletions  (file deletes)
src/MEDS_extract/download/source.py     | 609 / 70 additions
src/MEDS_extract/download/__init__.py   | minor
src/MEDS_extract/download/{cli,dispatch,backends/*}.py | minor
tests/{test_download,_fetcher_sigint_child}.py | minor
```

Net: **−111 lines**, two fewer files (`fetcher.py` and `_types.py` from the prior version of this PR both delete).

## Notes

- The four other open review items on PR #96 (drop `TGZ` enum, unify zip/tar via adapter, top-level imports, HIRID-specific phrasing) are still valid and can be folded in either here or after this lands. They're orthogonal to the architectural change.
- Targets `feat/fetcher-unarchive` (PR #96's branch) so the diff is reviewable as a clean delta. Merge order: this into `feat/fetcher-unarchive`, then retarget `feat/fetcher-unarchive` to `dev` and merge.


mmcdermott:
## Updated: dropped the `Fetcher` class

Reviewing the diff, you (rightly) flagged that `Fetcher` was a parameter object pretending to be a class. Its three responsibilities split cleanly:

- **Concurrency** is a property of a `ThreadPoolExecutor`, not a `Fetcher`.
- **Skip / overwrite / continue-on-error** are configuration → `DownloadPolicy` (frozen dataclass, no methods).
- **Orchestration** (submit work, collect results, run skip-and-fetch per file) is a free function (`_drive`) that takes a source, a pool, and a policy.

So `Fetcher` deletes. `Source.download_all` becomes:

```python
def download_all(self, dest_dir, *, pool=None, policy=None) -> FetchReport: ...
```

When `pool=None`, `download_all` builds a private 4-worker pool with the SIGINT-safe `cancel_futures` shutdown and tears it down on exit. When the caller passes a pool, the caller owns its lifetime — used by the CLI to share one pool across every source in a MESSY block via `ExitStack`. That's the shared-pool benefit we'd been talking about but the previous `Fetcher` class never realized (since `Fetcher.fetch_all` rebuilt the pool on every call).

### Cycle is gone, so `_types.py` deletes too

The previous version of this PR had a `_types.py` to break the source ↔ fetcher import cycle (caused by `Source.__init__`'s `Fetcher()` default). With `Fetcher` gone, no cycle. Everything lives in `source.py`:

- `RemoteFile` (manifest row, internal)
- `ChecksumError`, `sha256_of` (transport helpers)
- `DownloadPolicy` (frozen config)
- `FetchResult`, `FetchReport` (outcome types)
- `_resolve_dest`, `_already_complete`, `_maybe_unarchive`, `_fetch_one`, `_drive` (free functions, module-private)
- `Source` ABC

One file, ~580 lines, no internal cycle, no third module.

### Net delta vs the previous version of this PR

**−111 lines**, two fewer files (`fetcher.py` and `_types.py` both delete), one fewer class in the public API (`DownloadPolicy` replaces `Fetcher`). All 126 tests still pass; pre-commit clean.

PR description updated to match. Force-push not needed — the changes are on top of the previous redesign as a separate commit (`7e23763 Drop Fetcher class; pool-as-arg + DownloadPolicy on download_all`), so the review history shows the evolution.

PR #103, opened 2026-04-27
Redesign download API: Source.download_all + Source/RemoteFile

mmcdermott:
## Summary

Architectural redesign of the download module — extracted from #96 so it can be reviewed independently of the unarchive feature.

- **`Source.download_all(dest, *, pool=None, continue_on_error=False, do_overwrite=False)`** is the single public fetch entry point. The `Fetcher` class is gone; `continue_on_error` / `do_overwrite` are plain kwargs.
- **`Source` ABC** — backends implement two hooks (`_list_files`, `_pull`); the base class supplies `download_all`, the `files` cached+validated manifest property (with generic `include:`/`exclude:` fnmatch filtering), the per-file skip/overwrite/error policy, `.part` staging + SHA-256 verification + atomic rename, and the sequential/parallel orchestration.
- **`RemoteFile`** is a frozen, self-validating dataclass (`rel_path`, `source_path`, `sha256`) — every `_list_files` implementation constructs it, so it is exported; malformed rows (traversal, absolute paths, backslashes, non-64-hex digests) fail at construction.
- **Sequential by default**: `download_all` runs in the calling thread; pass an `Executor` to parallelize. The caller owns the pool. No implicit pool building inside the library.
- **HTTP transport**: both manifest GETs and streaming downloads retry transient failures (timeouts, connect errors, TCP resets, 5xx) with backoff; mid-body failures Range-resume from the enlarged `.part` on the retried attempt; `Accept-Encoding: identity` keeps on-wire bytes, resume offsets, and manifest digests in one byte space; a complete leftover `.part` is promoted without re-fetching.
- **CLI** builds one shared `ThreadPoolExecutor` via `ExitStack` and hands it to every source's `download_all`, validates `key=` against the spec's declared buckets, rejects cross-source destination collisions up-front (`validate_unique_destinations`), fails fast across sources unless `continue_on_error=true`, and **exits non-zero on failure** (explicit `sys.exit` — Hydra discards task-function return values).
- **Strict overwrite semantics**: an existing `dest` that can't be verified against the manifest's `sha256` is a hard `FileExistsError`, not a silent overwrite. Only a verified match is skipped; only `do_overwrite=True` overwrites.
- **Packaging**: `spec.py` (née `dispatch.py`) maps `type:` → backend via one lazy registry; `FsspecSource`, the `Source` ABC, and the CLI import and run without the `[download]` extra — only `HTTPSource`/`PhysioNetSource` need `httpx`/`tenacity`.
[...]

PR #104, opened 2026-04-27
Add post-fetch archive extraction (#92)

### LINKED ISSUE #92: Fetcher: optional post-fetch archive unpack (zip, tar.gz)

mmcdermott:
## Summary

Adds optional post-fetch archive extraction to the download layer. Stacks on top of #103 (download API redesign) so this PR is a clean delta of the unarchive feature only.

- **``unarchive.py``** — new module. ``ArchiveFormat`` (StrEnum) + ``safe_extract`` with format dispatch. Path-traversal validation pre-pass on every member; tar branch additionally uses PEP 706 ``data_filter`` as defense-in-depth. Stdlib zipfile/tarfile only — no third-party archive deps.
- **``RemoteFile``** gains ``unarchive`` and tri-state ``cleanup_archive`` fields. Both default to ``None`` (no unpack). ``"auto"`` mode infers the format from the file extension and is a no-op on non-archive files; explicit values (``zip``, ``tar``, ``tar.gz``, ``tgz``) override the inference.
- **Orchestration** runs a ``_maybe_unarchive`` hook in ``_fetch_one`` after the transport's ``_fetch`` succeeds. Tri-state cleanup: ``cleanup_archive=None`` defers to mode (``auto`` removes, explicit keeps); ``True``/``False`` always wins.
- **Backends**: ``HTTPSource`` accepts per-entry ``unarchive``/``cleanup_archive`` (motivating case: AUMCdb's DataVerse zip). ``PhysioNetSource`` accepts source-level kwargs that propagate to every listed member (motivating case: HIRID releases that mix archive and non-archive members under one SHA256SUMS).

This replaces #96, which had become entangled with the architectural redesign. The redesign now lives in #103; this PR is purely the archive feature on top of that.

Closes #92
Refs #96 #103

## Test plan

- [x] Unarchive doctests cover zip + tar.gz round-trip, zip-slip / tar-slip / absolute-path / link-target rejection (target dir empty after rejection — fail-closed verified), and ``AUTO`` / ``TGZ`` alias resolution.
- [x] ``HTTPSource._list_files`` and ``PhysioNetSource._list_files`` doctests verify the unarchive fields propagate from per-entry / source-level kwargs to ``RemoteFile``.
- [x] Full test suite green (129 tests, 2 deselected for ``integration`` marker — 5 more than #103's baseline; the deltas are the new unarchive doctests).
- [ ] CI green on push.


src/MEDS_extract/download/unarchive.py
```diff
@@ -0,0 +1,439 @@
...
+            "member path would land outside the extraction directory."
+        )
+    if link_target is not None and not _is_safe_path(link_target, target_root):
+        raise ValueError(
+            f"Refusing to extract unsafe {fmt_label} member {member_name!r} from {archive_path}: "
+            f"link target {link_target!r} would land outside the extraction directory."
+        )
+
+
+def _is_safe_path(member_name: str, target_root: Path) -> bool:
```
mmcdermott, line 253: Can some of this be now deduped with the other download path safety code?

mmcdermott: Good catch — done in 55d2615.

`Source._resolve_dest` and `_is_safe_path` were both answering the same question ("does this relative path, resolved on a real filesystem, stay inside this root?") with their own resolve + `relative_to`. Two copies of one security boundary is where drift actually bites, so they now share `download/_paths.resolve_contained`.

It returns `(resolved, is_contained)` instead of raising, because what genuinely differs between the callers isn't the check — it's that archive members are POSIX-shaped by format (a zip's `a/b.txt` means that on Windows too, hence `posix_member=True`) and each caller words its own error. Returning the resolved path keeps both messages as specific as they were.

New module rather than putting it in either file: `source.py` already imports `unarchive`, so anywhere else would be a cycle.

**One thing I deliberately left alone**, and noted in the module docstring so nobody "finishes the job" later: `RemoteFile.__post_init__`'s check. It looks like a third copy but isn't — it inspects the manifest string at construction, before any filesystem exists (backslashes, `..`, absolute). The shared one is the fetch-time check for escapes only a real directory can produce, symlinks especially. Collapsing those two would lose the construction-time rejection.

PR #126, opened 2026-07-10
dev: null code components as an explicit author choice (dftly 0.5 `??`)

mmcdermott:
Makes null `code` components an **explicit author choice** on the 0.7 line, using dftly 0.5's new `??` coalesce operator — reverting the automatic `"UNK"` fill from #118. This intentionally diverges from main (0.6.x), where null components are silently auto-filled with `"UNK"`.

## Behavior change
A composite `code` now uses the **native** dftly expression, which null-propagates: if any interpolated component is null, the whole `code` is null and the row is **dropped** (a MEDS `code` may never be null). The author decides, **per component**, whether a missing value drops the row or is filled in — by coalescing with `??`:

| MESSY `code` | null component → |
|---|---|
| `f"{$itemid}//{$valueuom}"` | row dropped |
| `f"{$itemid ?? 'UNK'}//{$valueuom ?? 'UNK'}"` | filled with `UNK`, row kept |
| `f"{$itemid}//{$valueuom ?? 'UNK'}"` | unit filled; a null `itemid` still drops |

(Single-quote the literal fallback *inside* the double-quoted f-string; `?? "UNK"` with double quotes clashes with the f-string delimiter.)

## Changes
- **`pyproject.toml` / `uv.lock`**: `dftly>=0.3.0` → `>=0.5.0` (0.5.0 introduces `??`).
- **`config.py`**: remove `_null_safe_code_expr`; `code` = native dftly expr; the null-code drop filter now keys on the **code expression** being non-null (so `??`-rescued rows survive), replacing the bare-column-only drop.
- **`config.py` docstring**: rewritten null-handling section (plain → drop; `??` → keep).
- **`README.md`** → *Code Construction* gains a **"Null components in composite codes"** section with runnable doctests for all three options above.

## Testing
- Full suite **74 passed** (1 integration deselected); config.py + README doctests pass under dftly 0.5.0.
- `ruff`, `mdformat`, `codespell` clean.

## Notes
- Refs #121 (design discussion of the desired behavior and user control) — this is the dev direction: user-controlled, not a hidden default.
- The `??` operator is dftly's resolution of the null-safe-interpolation request (dftly#75).

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,61 @@ from the parser:
 
 ```
 
+#### Null components in composite codes
+
+A MEDS `code` may never be null, and string interpolation **null-propagates**: if any interpolated
+component is null, the whole `code` becomes null and that row is **dropped**. To keep such rows, give
+the component a fallback with dftly's `??` (coalesce) operator — single-quote a literal fallback
+*inside* the double-quoted f-string. Whether a missing component drops the row or is filled in is
+your choice, made per component:
+
+```python
+>>> import polars as pl
+>>> from MEDS_extract.config import EventConfig
+>>> labs = pl.DataFrame({
+...     "subject_id": [1, 2, 3, 4],
+...     "itemid": ["GLU", "GLU", None, None],        # present, present, null, null
+...     "valueuom": ["mg/dL", None, "mg/dL", None],  # present, null,    present, null
+... })
+>>> def codes(expr):  # surviving (subject_id, code) rows for a `code` expression
+...     ev = EventConfig.parse("lab", {"code": expr, "time": None})
+...     return ev.extract(labs.lazy(), "labs/lab").collect().sort("subject_id").select(
+...         "subject_id", "code"
+...     )
+>>> codes('f"{$itemid}//{$valueuom}"')  # plain: any null component drops the row
+shape: (1, 2)
+┌────────────┬────────────┐
+│ subject_id ┆ code       │
+│ ---        ┆ ---        │
+│ i64        ┆ str        │
+╞════════════╪════════════╡
+│ 1          ┆ GLU//mg/dL │
+└────────────┴────────────┘
+>>> codes("f\"{$itemid ?? 'UNK'}//{$valueuom ?? 'UNK'}\"")  # fill both -> keep every row
+shape: (4, 2)
+┌────────────┬────────────┐
+│ subject_id ┆ code       │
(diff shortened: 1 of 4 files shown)


src/MEDS_extract/config.py
```diff
@@ -383,35 +363,33 @@ def extract(
...
-            │ 3          ┆ UNK//mg/dL │
-            │ 4          ┆ UNK//UNK   │
             └────────────┴────────────┘
 
+            To keep rows with missing components, coalesce them in the MESSY expression with the
+            dftly ``??`` operator (single-quote the literal fallback inside the f-string) — e.g.
+            ``code: f"{$itemid ?? 'UNK'}//{$valueuom ?? 'UNK'}"`` fills each null component with
+            ``"UNK"`` (yielding ``GLU//UNK``, ``UNK//mg/dL``, ``UNK//UNK``) so all four rows are
+            retained. See the README's *Code Construction* section for runnable examples of each
+            option.
```
mmcdermott: I want an explicit doctest here showing the capability.

README.md
```diff
@@ -437,6 +437,61 @@ from the parser:
...
+│ subject_id ┆ code       │
+│ ---        ┆ ---        │
+│ i64        ┆ str        │
+╞════════════╪════════════╡
+│ 1          ┆ GLU//mg/dL │
+│ 2          ┆ GLU//UNK   │
+│ 3          ┆ UNK//mg/dL │
+│ 4          ┆ UNK//UNK   │
+└────────────┴────────────┘
+>>> codes("f\"{$itemid}//{$valueuom ?? 'UNK'}\"")  # fill only the unit: a null itemid still drops
```
mmcdermott, line 494: Can you have one example where the interpolated code isn't using UNK but a different string, to show it is possible?

src/MEDS_extract/config.py
mmcdermott: Done — the passage now carries a runnable doctest: the four-row null-components frame from the preceding example, extracted with `'''f"{$itemid ?? 'UNK'}//{$valueuom ?? 'UNK'}"'''`, showing all four rows retained with the filled codes.

README.md
mmcdermott: Done — added a `?? 'NO_ITEM'` / `?? 'NO_UNIT'` example to the runnable README block, showing the fallback is any literal the author chooses rather than a magic `UNK`.

PR #139, opened 2026-07-23
extract_code_metadata: fix partial-match component-join correctness (#110, #134, #135)

mmcdermott:
## Summary

Fixes three verified defects in the partial-match (`_match_on`) component-join region of `extract_code_metadata`, which the issues flagged as needing to land together:

- **#110 — `DuplicateError` on a source column named `code` (+ silent misclassification):** the code-component map now keeps the fully-assembled output code under the reserved internal alias `__meds_full_code`, so unnesting the `code_components` struct can never collide with an output `code` column — even when a code expression references a source column literally named `code` (`code: 'f"ICD//{$code}"'`). The reducer's full-vs-partial classification is now driven by explicit map-time bookkeeping (`out_fp -> (match_cols, source_block)`) instead of sniffing output schemas for a `code` column, which misclassified a partial output whose `_match_on` includes `code` as full-match and passed raw component values through as output codes.
- **#134 — event-unscoped `_match_on` join:** `MessyConfig.events_by_metadata_prefix` now tags each entry with the declaring event's `source_block` (`{input_prefix}/{event_name}`, the same tag `EventConfig.extract` already stamps on every data row), the component map carries `source_block`, and each partial-match expansion filters the map to the declaring event before joining — so same-named component columns with colliding values on *other* events no longer receive the metadata or a false `code_template` provenance stamp. The expansion also selects the code alias + metadata columns explicitly, fixing the verified name-collision variant where a metadata column sharing a component column's name silently yielded the component values.
- **#135 — dtype-mismatch `SchemaError`:** partial-match join keys are normalized to one canonical String rendering on both sides via the new `normalize_join_key` helper (doctested): typed `Int64` components join all-String (`infer_schema=False`) CSV metadata keys, and integer-valued float components render as `220045`, not `220045.0` (via a lossless `Int64` cast; integer-valued floats outside `Int64` range render as null and match nothing, as documented). A WARNING is emitted when a partial join matches zero codes — the full match-coverage diagnostics remain tracked in #138.

Closes #110
Closes #134
Closes #135

**Supersedes and absorbs #114:** this branch was cut from `tests/issue-110-metadata-dup-code-demo`, so it carries that PR's demonstration test — now green, with its strict xfail marker removed per the test's own instructions and the assertions strengthened to check the reduced metadata contents.

## Test plan

- [x] PR #114's `test_extract_code_metadata_handles_code_named_source_column` flipped from strict-xfail to passing (asserts stage completion and correct full-match metadata on reconstructed codes)
- [x] New: `_match_on: code` misclassification regression — partial output keyed on a column named `code` produces partial-match expansion (`ICD//250.00`), not full-match passthrough of raw component values
- [x] New: #134 scoping regression from the issue's verified repro — `CHART//{$itemid}` and `LAB//{$itemid}` sharing itemid values, `d_items` declared only on the chartevents event: `LAB//1` receives no metadata and no false `code_template`
- [x] New: #135 dtype regressions — typed `Int64` itemid components join CSV metadata correctly; integer-valued float components match `220045` (and non-integer floats match their float rendering)
- [x] New: zero-match WARNING emitted when a partial join matches no codes
- [x] `normalize_join_key` and updated `events_by_metadata_prefix` covered by doctests
- [x] Full suite green: `uv run pytest -q` → 85 passed, 1 deselected
- [x] `uv run pre-commit run --all-files` stable (all hooks pass)

### DIFF
--- a/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
+++ b/src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
@@ -32,29 +32,30 @@ MEDS_METADATA_MANDATORY_TYPES = {
 
 # Reserved internal alias for the fully-assembled output code inside the code-component map.
 # Keeping the full code under this name (rather than "code") means unnesting the
-# ``code_components`` struct can never collide with it — even when a code expression
-# references a source column literally named ``code`` (see issue #110).
+# ``code_components`` struct can never collide with it — a code expression may reference
+# a source column literally named ``code`` (the idiomatic ICD/OMOP vocabulary shape).
 FULL_CODE_COL = "__meds_full_code"
 
 
-def normalize_join_key(col: str, dtype: pl.DataType) -> pl.Expr:
-    """Render the join-key column ``col`` of type ``dtype`` as a canonical String expression.
+def normalize_join_key(expr: pl.Expr, dtype: pl.DataType) -> pl.Expr:
+    """Render the join-key expression ``expr`` of type ``dtype`` as a canonical String expression.
 
     Code components keep their raw source dtypes (an ``Int64`` ``itemid``, say), while
     csv-sourced metadata keys are uniformly ``String`` (they are read with
-    ``infer_schema=False``). Joining the two directly raises a ``SchemaError`` (issue #135),
-    so both sides of the partial-match join are normalized through this single canonical
-    string rendering:
+    ``infer_schema=False``). Joining the two directly is a ``SchemaError``, so both sides
+    of the partial-match join are normalized through this single canonical string
+    rendering:
 
-    - String columns pass through unchanged.
-    - Non-float columns cast directly: ``220045`` renders as ``"220045"``.
-    - Float columns render integer-valued entries via ``Int64`` so that ``220045.0`` matches
-      the metadata string ``"220045"`` rather than rendering as ``"220045.0"``; non-integer
-      values keep their float rendering (``1.5`` renders as ``"1.5"``). Integer-valued floats
-      outside the ``Int64`` range render as null (and thus match nothing).
+    - String expressions pass through unchanged.
+    - Non-float expressions cast directly: ``220045`` renders as ``"220045"``.
+    - Float expressions render integer-valued entries via ``Int64`` so that ``220045.0``
+      matches the metadata string ``"220045"`` rather than rendering as ``"220045.0"``;
+      non-integer values keep their float rendering (``1.5`` renders as ``"1.5"``).
+      Integer-valued floats outside the ``Int64`` range render as null (and thus match
+      nothing).
 
-    The returned expression preserves the column name (nulls stay null and never match under
-    the default ``join_nulls=False``).
+    The output keeps the input expression's root name (nulls stay null and never match
+    under the default ``join_nulls=False``).
 
     Examples:
         >>> df = pl.DataFrame({
@@ -63,9 +64,9 @@ def normalize_join_key(col: str, dtype: pl.DataType) -> pl.Expr:
         ...     "s": ["220045", "01.90", None],
         ... })
         >>> df.select(
-        ...     normalize_join_key("i", df.schema["i"]),
-        ...     normalize_join_key("f", df.schema["f"]),
-        ...     normalize_join_key("s", df.schema["s"]),
+        ...     normalize_join_key(pl.col("i"), df.schema["i"]),
+        ...     normalize_join_key(pl.col("f"), df.schema["f"]),
+        ...     normalize_join_key(pl.col("s"), df.schema["s"]),
         ... )
         shape: (3, 3)
         ┌────────┬────────┬────────┐
@@ -78,7 +79,6 @@ def normalize_join_key(col: str, dtype: pl.DataType) -> pl.Expr:
         │ null   ┆ null   ┆ null   │
         └────────┴────────┴────────┘
     """
-    expr = pl.col(col)
     if dtype == pl.String:
         return expr
     if dtype.is_float():
@@ -86,9 +86,51 @@ def normalize_join_key(col: str, dtype: pl.DataType) -> pl.Expr:
             pl.when(expr == expr.round(0))
             .then(expr.cast(pl.Int64, strict=False).cast(pl.String))
             .otherwise(expr.cast(pl.String))
-            .alias(col)
+            .name.keep()
         )
     return expr.cast(pl.String)
+
+
+def validate_event_data_schema(data_schema: pl.Schema) -> bool:
+    """Validate extracted-event input schema for metadata extraction; return whether components exist.
+
+    ``code_components`` is only attached by ``EventConfig.extract`` when a code expression
+    references at least one source column — a dataset whose codes are all literals
+    legitimately has no components (and therefore nothing to partial-match on), so its
+    absence is allowed and reported as ``False``.
+
+    When components ARE present, ``source_block`` must be too: ``EventConfig.extract``
+    stamps it on every row unconditionally, so its absence means the events were produced
+    by a pre-0.7 extraction pipeline — and without it, partial-match metadata cannot be
+    scoped to the event that declared it (one event's metadata would silently attach to
+    other events' codes sharing a component value).
+
+    Examples:
+        >>> validate_event_data_schema(pl.Schema({"code": pl.String}))
+        False
+        >>> validate_event_data_schema(pl.Schema({
+        ...     "code": pl.String,
+        ...     "code_components": pl.Struct({"itemid": pl.Int64}),
+        ...     "source_block": pl.String,
+        ... }))
+        True
+        >>> validate_event_data_schema(pl.Schema({
+        ...     "code": pl.String,
+        ...     "code_components": pl.Struct({"itemid": pl.Int64}),
+        ... }))
+        Traceback (most recent call last):
+            ...
+        ValueError: Extracted event data carries 'code_components' but no 'source_block' column. ...
+    """
+    if "code_components" not in data_schema:
+        return False
+    if SOURCE_BLOCK_COL not in data_schema:
+        raise ValueError(
+            f"Extracted event data carries 'code_components' but no {SOURCE_BLOCK_COL!r} "
+            "column. These events were produced by a pre-0.7 convert_to_MEDS_events; "
+            "re-run the extraction pipeline before extracting code metadata."
+        )
+    return True
 
 
 def extract_metadata(
@@ -467,22 +509,10 @@ def main(cfg: DictConfig):
     all_data = pl.concat(all_event_dfs, how="diagonal_relaxed")
     all_codes = all_data.select(pl.col("code").unique()).collect().get_column("code").to_list()
 
-    # Build code_components mapping for partial metadata joins. The full output code is kept
-    # under the reserved FULL_CODE_COL alias so that unnesting the struct can never collide
-    # with it — a code expression may reference a source column literally named "code" (#110).
-    # source_block is carried along so each partial-match expansion can be scoped to the event
-    # that declared the _metadata block (#134).
-    data_schema = all_data.collect_schema()
-    if "code_components" in data_schema:
-        if SOURCE_BLOCK_COL not in data_schema:
-            # ``EventConfig.extract`` stamps source_block on every row unconditionally, so
-            # its absence means the events predate 0.7's extraction pipeline — partial-match
-            # scoping (#134) cannot work without it.
-            raise ValueError(
-                f"Extracted event data carries 'code_components' but no {SOURCE_BLOCK_COL!r} "
-                "column. These events were produced by a pre-0.7 convert_to_MEDS_events; "
-                "re-run the extraction pipeline before extracting code metadata."
-            )
+    # Build the code_components mapping for partial metadata joins: full code (under the
+    # reserved collision-proof alias), unnested component columns, and the declaring
+    # source_block so each expansion joins only against its own event's codes.
+    if validate_event_data_schema(all_data.collect_schema()):
         code_component_map = (
             all_data.select(pl.col("code").alias(FULL_CODE_COL), "code_components", SOURCE_BLOCK_COL)
             .unique()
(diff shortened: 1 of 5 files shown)


src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -407,18 +467,26 @@ def main(cfg: DictConfig):
...
+    # with it — a code expression may reference a source column literally named "code" (#110).
+    # source_block is carried along so each partial-match expansion can be scoped to the event
+    # that declared the _metadata block (#134).
     data_schema = all_data.collect_schema()
     if "code_components" in data_schema:
-        code_component_map = (
-            all_data.select("code", "code_components").unique().collect().unnest("code_components")
-        )
+        map_cols = [pl.col("code").alias(FULL_CODE_COL), pl.col("code_components")]
+        if "source_block" in data_schema:
```
mmcdermott: if "source_block" is being used across files or locations, it should be made into a named constant somewhere.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -30,6 +30,66 @@
...
 }
 
+# Reserved internal alias for the fully-assembled output code inside the code-component map.
+# Keeping the full code under this name (rather than "code") means unnesting the
+# ``code_components`` struct can never collide with it — even when a code expression
+# references a source column literally named ``code`` (see issue #110).
+FULL_CODE_COL = "__meds_full_code"
+
+
+def normalize_join_key(col: str, dtype: pl.DataType) -> pl.Expr:
```
mmcdermott: Re issue 135, it seems like an easier and better solution would be to make the code components struct preserve the raw source types, no? Or is that not possible given how the struct is constructed or how data is read for some reason?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: and should this be `"source_block"` or `"source_table_name"` or something, as it does map to the actual source table, right?

mmcdermott: Also, should this _always_ be present? When would it not be present? 0.7.0 will be a breaking change on many axes, so it is possible to introduce something mandatory and new.

mmcdermott: Also, perhaps this should be integrated with the provenance data / effort somehow? That can be reserved for a later deduplication though, or provenance can be added first if that change is simpler.

tests/test_issue_110_metadata_code_column.py
```diff
@@ -0,0 +1,160 @@
...
+    ecm_stage.main_fn(cfg)
+    return out_dir
+
+
+def test_convert_produces_code_components_with_a_code_field():
+    """Provenance: the real convert stage attaches a ``code_components`` struct whose field
+    is literally ``code`` when the code references a ``$code`` source column — this is what
+    makes the ``extract_code_metadata`` unnest collide.
+    """
+    raw = pl.DataFrame({"subject_id": [1], "code": ["250.00"], "ts": ["2020-01-01"]})
```
mmcdermott: This should be a doctest.

tests/test_extract_code_metadata.py
mmcdermott, line 1: This test file references an issue by name when that name that will lose relevance after the issue is solved. Can you rename this test file (and potentially split this and other tests) into more semantic units?

tests/test_issue_110_metadata_code_column.py
```diff
@@ -0,0 +1,160 @@
...
+        }
+    )
+    ecm_stage.main_fn(cfg)
+    return out_dir
+
+
+def test_convert_produces_code_components_with_a_code_field():
+    """Provenance: the real convert stage attaches a ``code_components`` struct whose field
+    is literally ``code`` when the code references a ``$code`` source column — this is what
+    makes the ``extract_code_metadata`` unnest collide.
```
mmcdermott: Should be a doctest

mmcdermott (commented):
Some general comments but overall I think this is in the right direction. Strongly consider though using https://github.com/mmcdermott/yaml_to_disk and https://github.com/Medical-Event-Data-Standard/meds_testing_helpers and https://github.com/mmcdermott/pretty-print-directory to simply testing code and move things into readable doctests where appropriate.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Done in 8d8bd84 — `SOURCE_BLOCK_COL` is a documented module constant in config.py (where `EventConfig.extract` stamps it), imported and used everywhere `extract_code_metadata.py` touches the column. Merged forward through the stack (#140, #141), all green (84/87/140 tests).

mmcdermott: I'd push back gently on `source_table_name`: the value is `f"{table_prefix}/{event_name}"` — one *event block*, finer-grained than a source table (two events in the same table carry distinct values, which is exactly the granularity #134's scoping needs — `d_items` declared on `chartevents/hr` must not attach to `chartevents/bp` codes either). `source_block` names the MESSY config block, which is what it is; the new constant's comment in 8d8bd84 spells out the granularity so the name can't mislead. Happy to rename to something like `source_event_block` if you want the granularity in the name itself.

mmcdermott: Agreed, and done in 8d8bd84 under the 0.7.0 breaking-change license: `EventConfig.extract` already stamps it unconditionally on every row, so the only absent case is pre-0.7 extracted data. The stage now (a) raises a clear ValueError ("produced by a pre-0.7 convert_to_MEDS_events; re-run the extraction pipeline") when `code_components` exists without `source_block`, (b) `pop`s the per-entry `source_block` with no fallback, and (c) applies the scoping filter unconditionally — the `str | None` typing and the conditional fallback paths are gone.

mmcdermott: Agreed on deferring, and I'd frame them as complementary layers rather than duplicates: `source_block` is *config* lineage (which MESSY block produced the row — always-on, cheap, and load-bearing for metadata scoping), while #132/#133's `provenance` is *data* lineage (which input file+row — opt-in, list-valued, merged at dedup points). #133 deliberately keeps `source_block` as its own column rather than folding it into the struct, so nothing blocks either landing first. I've noted the potential later unification (e.g. `source_block` as a provenance-struct field once provenance is always-on) on #132 so it isn't lost.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: The struct actually already preserves the raw source dtypes — that's precisely what *creates* #135: the metadata side is deliberately read `infer_schema=False`, because schema inference is at its worst on vocabulary tables (zero-padded ICD codes like `01.90` infer to floats, leading-zero identifiers corrupt, mixed alpha-numeric columns infer differently per chunk). So the two sides genuinely arrive typed-vs-String, and the join has to pick a shared representation: inferring the metadata side would corrupt keys; casting metadata → component dtype is the lossy direction (and breaks when one metadata column serves differently-typed keys across events); so `normalize_join_key` renders both sides to canonical strings. That also matches the semantics — matching ultimately happens in code-rendering space, since the code string embeds the *rendered* component. And note typed-both-sides wouldn't dodge representation drift anyway (a Float `220045.0` component vs an Int-derived `220045` key still mismatches; the canonical rendering handles exactly that). Keeping the struct typed is also what #138 wants for exporting faithful component values.

tests/test_issue_110_metadata_code_column.py
mmcdermott: Done in 8d8bd84 — the struct-shape check is now a doctest on `EventConfig.extract` itself (where the behavior lives), and the pytest test is removed.

tests/test_issue_110_metadata_code_column.py
mmcdermott: Done in 8d8bd84 — converted to an `EventConfig.extract` doctest (same commit as the sibling thread).

tests/test_extract_code_metadata.py
mmcdermott: Done in 8d8bd84 — renamed to `tests/test_extract_code_metadata.py` (git mv, history preserved) with a semantic module docstring; issue numbers survive as reference links, not names. On splitting further: the natural companion move is relocating the `_run_ecm_scenario` stage-scenario tests out of `test_inprocess_pipeline.py` into this file — but #140 and #141 both extend that helper on stacked branches, so doing it mid-stack would conflict twice. Proposal: consolidate at the top of the stack (on #141) or as an immediate follow-up once the stack merges — say which you prefer.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -407,18 +467,37 @@ def main(cfg: DictConfig):
...
+    # under the reserved FULL_CODE_COL alias so that unnesting the struct can never collide
+    # with it — a code expression may reference a source column literally named "code" (#110).
+    # source_block is carried along so each partial-match expansion can be scoped to the event
+    # that declared the _metadata block (#134).
     data_schema = all_data.collect_schema()
     if "code_components" in data_schema:
+        if SOURCE_BLOCK_COL not in data_schema:
+            # ``EventConfig.extract`` stamps source_block on every row unconditionally, so
+            # its absence means the events predate 0.7's extraction pipeline — partial-match
+            # scoping (#134) cannot work without it.
```
mmcdermott: Referencing issue #s that will become stale is not a good practice.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -407,18 +467,37 @@ def main(cfg: DictConfig):
...
 
-    # Build code_components mapping for partial metadata joins
+    # Build code_components mapping for partial metadata joins. The full output code is kept
+    # under the reserved FULL_CODE_COL alias so that unnesting the struct can never collide
+    # with it — a code expression may reference a source column literally named "code" (#110).
+    # source_block is carried along so each partial-match expansion can be scoped to the event
+    # that declared the _metadata block (#134).
     data_schema = all_data.collect_schema()
     if "code_components" in data_schema:
+        if SOURCE_BLOCK_COL not in data_schema:
```
mmcdermott: Can you move this and the other validation of "code_components" up to occur earlier at the top of the function, so it is like a generic validation step? Or have a validate input helper or something that you can just call to move this logic out and test it via doctests directly?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: can this just take an expression instead of a col? That's what it really operates on, right? And you can use .name.keep() I think instead of the alias to the column name?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -407,18 +467,37 @@ def main(cfg: DictConfig):
...
     all_codes = all_data.select(pl.col("code").unique()).collect().get_column("code").to_list()
 
-    # Build code_components mapping for partial metadata joins
+    # Build code_components mapping for partial metadata joins. The full output code is kept
+    # under the reserved FULL_CODE_COL alias so that unnesting the struct can never collide
+    # with it — a code expression may reference a source column literally named "code" (#110).
+    # source_block is carried along so each partial-match expansion can be scoped to the event
+    # that declared the _metadata block (#134).
     data_schema = all_data.collect_schema()
     if "code_components" in data_schema:
```
mmcdermott: Why is this a condition? In what setting will this not be true?

mmcdermott: Also, should this be a constant?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Done in 91262cd — every comment this PR added now states its invariant self-containedly instead of citing an issue number (e.g. the FULL_CODE_COL note explains the vocabulary-shape collision directly). Pre-existing dev-era `#109` references in config.py were left as-is to keep the diff scoped; happy to sweep those separately.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Done in 91262cd — hoisted into a doctested module-level helper, `validate_event_data_schema(data_schema) -> bool`, called before the map build. Its doctests cover all three cases: no components (returns False), components + source_block (True), components without source_block (the pre-0.7 ValueError).

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Done in 91262cd — `normalize_join_key(expr: pl.Expr, dtype: pl.DataType)` now takes the expression it operates on, and `.name.keep()` replaces the alias round-trip; call sites pass `pl.col(c)`. (dtype stays a parameter since a lazy expression can't be introspected for type without a schema in hand.)

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Legitimately conditional: `EventConfig.extract` only attaches `code_components` when the code expression references at least one source column — a dataset whose codes are all literals (`code: "EYE_COLOR"`) has no components and therefore nothing to partial-match on. That rationale now lives in the doctested `validate_event_data_schema` helper's docstring (91262cd) rather than as an unexplained `if`. (Under the one-join-path unification we discussed, this becomes the explicit "literal-only codes have no component join" degenerate case.)

PR #145, opened 2026-07-24
extract_code_metadata: deterministic, stable-schema reduction (#137)

mmcdermott:
## Summary

Successor to #141 (auto-closed when its stacked base branch was deleted; GitHub cannot reopen a closed PR after a force-push). The four fix areas from the reducer-correctness investigation, rebased directly onto dev (post-#126, post-#139), updated per review:

1. **Deterministic reduction** — partial/full frames reduce in canonical `(metadata_prefix, cfg_idx)` order regardless of worker shuffle; aggregations are order-stable; output sorted by code → byte-identical `codes.parquet` across runs.
2. **Data-independent, deduplicated output schema** — values are deduplicated everywhere (identical metadata repeated across sources collapses): `description` joins **distinct** values with `description_separator` in config order; `parent_codes` is `List(String)`, deduplicated in first-seen order; extra metadata columns are always `List(String)` of distinct values, sorted; missing values are null, never `""`/`[]`.
3. **`code_template` is a plain `String`** — one code has exactly one template; distinct templates colliding on one code raise a config error naming the offenders. `code` and `code_template` are validated as **reserved** `_metadata` output names (pipeline-generated; users cannot redefine them).
4. **Crashers**: mixed csv+parquet metadata prefixes work (read kwargs chosen per file; was `TypeError`). The accidental `_match_on` key-rename capability (join key doubling as a `_metadata` output expression) is **removed, not fixed** — it now raises a clear `ValueError`; key sourcing/renaming belongs to the planned dftly metadata-block design (tracking issue linked below).
5. **Pre-existing `codes.parquet` merge** coalesces same-named columns (extracted wins; dtype conflicts raise naming the column) instead of silently forking `*_right` duplicates.

Closes #137. Refs #144 (unification builds on this reducer; the dftly metadata-block proposal extends it), #126.

## Test plan

- [x] Full suite green locally: 139 passed, 2 deselected (integration).
- [x] Regression tests: determinism across forced opposite shuffle orders (byte-identity), data-independent schema, value deduplication, null-not-empty, merge coalescing + dtype-conflict error, template-conflict error, reserved-name and alias `ValueError`s (doctested), mixed-format prefix.
- [x] `pre-commit run --all-files` stable.
- [ ] Full CI green (dev-based PR — test + code-quality workflows fire directly).


src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -590,9 +641,13 @@ def read_fn(fps, _kwargs=read_kwargs):
...
     # whose _match_on includes "code" carries a "code" column of raw component values and
     # would be silently misclassified as full-match by schema sniffing.
+    #
+    # Frames are processed in canonical (metadata_prefix, cfg_idx) order — NOT in this
+    # worker's shuffled map order — so the reduction (concat order, and with it description
+    # join order and list-aggregation order) is identical across runs and workers.
     full_match_dfs = []
     partial_match_dfs = []
-    for fp in all_out_fps:
+    for fp in sorted(all_out_fps, key=out_fp_keys.__getitem__):
```
mmcdermott, line 653: I think this is a problem. Are you doing this to solve an issue? If so, this is the wrong way to do it -- this will make all mappers try to read files in identical order (as they all get the same set of files to work over, there is no centralized queue management) so making orders preserved will dramatically increase the chance of collisions and make partial runtime estimates invalid. If you want to make them sorted in a particular way at output, you need to propagate it through the data and do it in a later step. I'm also not sure this is that important, and the bug can likely be deferred to another PR if it doesn't cause a high-value problem.

src/MEDS_extract/io.py
```diff
@@ -88,6 +90,33 @@ def scan_source(
...
+        >>> with yaml_disk('''
+        ... items/a.csv: |
+        ...   itemid,label
+        ...   1,Heart Rate
+        ... items/b.parquet:
+        ...   itemid: [2]
+        ...   label: [NBP systolic]
+        ... ''') as d:
+        ...     fps = sorted((Path(d) / 'items').glob('*'))
+        ...     scan_source(fps, infer_schema=False).collect().sort('itemid')
```
mmcdermott: In this setting, what happens if the parquet is typed and not a string? How does that work? E.g., if you add a "numeric value" column? I think we either want (1) to not allow cross type reads like this -- why do we need this again? -- or (2) to ensure that any parquet files have a consistent schema then cast csv outputs to that schema.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Pushing back on this one — the concern doesn't apply, but the comment invited the misread, so I've hardened it (92ef2ed; see the updated block). This sort is **reducer-only**: it runs on worker 0 alone, after `wait_for_complete_parquets`, and orders the list of *already-written* partial files for concatenation. The mappers' per-worker `random.shuffle` (line ~547) is untouched, so map-phase FileLock contention and partial-runtime spreading are exactly as before — this is already the "propagate through the data and order in a later step" shape. On deferring: the nondeterminism it fixes is user-visible (`codes.parquet` differing across identical runs — description join order and aggregation order followed shuffle order), and the fix is this one sort over a small in-memory list on one worker, so I'd keep it here.

src/MEDS_extract/io.py
mmcdermott: You're right — and the doctest itself betrayed the problem (`Int64` parquet + all-String csv silently unifying to `String` via `vertical_relaxed` — e.g. a numeric column would be stringified). Honest answer to "why do we need this": we don't; it was an adversarially-found crash (csv kwargs reaching `scan_parquet` → `TypeError`) and supporting the mix invented semantics. Implemented option (1): `scan_source` now rejects a multi-file source mixing csv- and parquet-family chunks with a clear `ValueError` ("convert the chunks to a single format"), doctested; the regression test asserts the error instead of the coercion. One nuance kept deliberately: **single-file** scans still tolerate csv-inference kwargs on parquet inputs — that's pre-existing dev behavior `shard_events` relies on (it passes one kwargs set while scanning raw files individually, whatever each file's format), and it's safe precisely because multi-file sources are now format-homogeneous. Option (2) (cast csv to a consistent parquet schema) remains available later if a real mixed-chunk use case ever appears — it would slot into the #143 strings-first typing story.

src/MEDS_extract/io.py
```diff
@@ -118,7 +159,12 @@ def _scan_one(fp: Path | UPath, **scan_kwargs: Any) -> pl.LazyFrame:
         # glob=False: we've already resolved the exact file path, so polars must
         # treat it literally. Critical for shard_events' "[0-10).parquet" output,
         # where the filename itself contains glob metacharacters.
+        # csv-inference kwargs are ignored for parquet: callers like shard_events pass
+        # one kwargs set while scanning raw files individually, whatever each file's
+        # format. (Multi-file scans are format-homogeneous — enforced in scan_source —
+        # so this never silently mixes typed and inferred chunks of one source.)
         scan_kwargs.pop("infer_schema_length", None)
+        scan_kwargs.pop("infer_schema", None)
```
mmcdermott: Is infer_schema ever used anymore? Do we still need it at all here as an arg, or elsewhere? What role does it play now?

mmcdermott: It splits three ways — addressed in 03d9658:

1. **This call site is `infer_schema`'s one live, load-bearing use**: `extract_code_metadata` passes `infer_schema=False` for csv-family metadata prefixes — the deliberate all-String vocabulary reads the raw-join policy rests on (zero-padded ICD codes, leading-zero itemids stay text). Under #143 (strings-first ingestion) this posture becomes the default for every CSV read, so its role grows rather than shrinks.
2. **Elsewhere it was dead**: the parquet-branch `pop` of `infer_schema` in `_scan_one` can no longer be reached — kwargs are chosen per resolved prefix (parquet prefixes get `{}`) and multi-file sources are now format-homogeneous — so 03d9658 removes it.
3. **`infer_schema_length` is the separate knob that keeps its parquet tolerance** — `shard_events` passes one kwargs set while scanning raw files individually, whatever each file's format (pre-existing; safe precisely because multi-file mixing now errors).

Also fixed in passing: `scan_source`'s docstring still claimed callers needn't care 'even when a single prefix mixes formats' — it now states the homogeneity requirement.

PR #162, opened 2026-07-29
RC-review cleanup bundle: dead code, cost gating, bounds, and API hygiene

mmcdermott:
Bundle of small code cleanups queued from the 0.7.0 RC design review, **one commit per finding** for easy per-item review (and easy dropping of any item you disagree with). Opened for maintainer review — not intended for auto-merge.

| Commit | Finding | Rationale |
|---|---|---|
| `9908e20` | Delete dead `EventConfig.time_source_columns` | Zero callers after #154 moved the computed-time null filter post-select; it only ever fed the deleted pre-filter derivation. |
| `0d32dbb` | Delete the dead `code_modifier_cols` reducer knob | Post-#147 the expansion join fully resolves metadata rows to assembled codes before reduction, so the reduced frame only ever carries a `code` key — any configured `code_modifier_cols` would *crash* the `n_unique`/`group_by` on a missing column. Never set by any config/doc/test. `join_cols` is now hardcoded to `["code"]`. |
| `61dc732` | Delete `MessyConfig.iter_events` | Its only reference was its own class doctest (now expressed directly over `cfg.tables`). `iter_tables` stays — it has a real caller in `split_and_shard_subjects`. |
| `092bd0b` | Gate the component-map `.collect()` behind the worker-0 reduction | Every worker materialized the code-component map — a full-dataset scan/unique/collect — but only worker 0's reduction consumes it; N−1 map-only workers paid the stage's most expensive non-map step for nothing. The cheap metadata-only schema validation stays pre-map in every worker so a pre-0.7 events layout still fails loudly everywhere. **Test**: `build_code_component_map` is monkeypatch-spied — a worker-1 run must not call it (nor write `codes.parquet`); a worker-0 run calls it exactly once and reduces normally. |
| `1499ddb` | pyproject: `dftly>=0.5.0,<0.6` + `universal_pathlib>=0.3.10` | dftly is pre-1.0 and the event layer leans on 0.5 contracts (`??` coalescing, `referenced_columns`/`NodeBase` internals); pin the verified minor. The unbounded `universal_pathlib` floor was a lie — floor at the lock-resolved 0.3.10 the suite actually verifies. |
| `16ff7f4` | Rename `_should_retry_stream` → `_should_retry` | The predicate is shared by both request paths (`_get` manifest GETs and `_pull` streaming); the `_stream` suffix misdescribes half its callers. Private; no external references. |
| `04d228f` | cli: log-and-exit on spec construction errors | `sources_from_spec` errors (unknown `type:`, bad backend kwargs) propagated raw through Hydra while manifest validation just below already had the friendly try/log/`sys.exit(1)` treatment. Same pattern applied; `TypeError` caught alongside `ValueError` because kwarg typos surface as constructor `TypeError`s. **Test**: `type: s3` spec exits non-zero with the CLI log line + the underlying "Unknown source type" message. |
| `9ce459f` | `extract_metadata` hygiene | (1) annotation drift: the function is laziness-preserving (its own doctests use eager frames), so annotate `pl.LazyFrame \| pl.DataFrame` both sides; `event_cfg` values are nested mappings, not `str \| None`. (2) the `isinstance` TypeError guard ran only *after* `deepcopy` had already accepted arbitrary junk — check first. (3) drop `extract_metadata`'s own `deepcopy` outright: it never mutates `event_cfg`, and `main()` already deepcopies each entry before popping `source_block` — the per-call deepcopy was duplicate work. |
| `06381cc` | docformatter reflow | Mechanical pre-commit reflow of two pre-existing docstrings in a touched file. |

Local: full suite green with extras (`159 passed, 2 deselected`); pre-commit clean over the whole branch diff.

Refs the 0.7.0 RC review.

### DIFF
--- a/src/MEDS_extract/download/cli.py
+++ b/src/MEDS_extract/download/cli.py
@@ -130,7 +130,14 @@ def main(cfg: DictConfig) -> None:
             if bucket_node is not None:
                 sources_dict[bucket] = OmegaConf.to_container(bucket_node, resolve=True)
 
-    sources = sources_from_spec({"sources": sources_dict}, key=cfg.key)
+    # Spec-shape errors (missing/unknown ``type:``, bad backend kwargs) are user config
+    # mistakes: log them and exit 1, mirroring the manifest-validation handling below,
+    # rather than dumping a raw traceback through Hydra.
+    try:
+        sources = sources_from_spec({"sources": sources_dict}, key=cfg.key)
+    except (TypeError, ValueError):
+        logger.exception(f"Could not construct sources from the spec at {spec_fp}")
+        sys.exit(1)
 
     if not sources:
         logger.warning(f"No sources resolved for key={cfg.key!r} in {spec_fp}. Nothing to do.")
--- a/src/MEDS_extract/download/cli.py
+++ b/src/MEDS_extract/download/cli.py
@@ -135,8 +135,8 @@ def main(cfg: DictConfig) -> None:
     # rather than dumping a raw traceback through Hydra.
     try:
         sources = sources_from_spec({"sources": sources_dict}, key=cfg.key)
-    except (TypeError, ValueError):
-        logger.exception(f"Could not construct sources from the spec at {spec_fp}")
+    except (TypeError, ValueError) as e:
+        logger.error(f"Could not construct sources from the spec at {spec_fp}: {e}")
         sys.exit(1)
 
     if not sources:
@@ -168,8 +168,8 @@ def main(cfg: DictConfig) -> None:
         # source-side hashing.)
         try:
             validate_unique_destinations(sources)
-        except ValueError:
-            logger.exception("Source manifests failed validation")
+        except ValueError as e:
+            logger.error(f"Source manifests failed validation: {e}")
             sys.exit(1)
 
         all_ok = True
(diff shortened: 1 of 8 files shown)


src/MEDS_extract/download/cli.py
```diff
@@ -130,7 +130,14 @@ def main(cfg: DictConfig) -> None:
             if bucket_node is not None:
                 sources_dict[bucket] = OmegaConf.to_container(bucket_node, resolve=True)
 
-    sources = sources_from_spec({"sources": sources_dict}, key=cfg.key)
+    # Spec-shape errors (missing/unknown ``type:``, bad backend kwargs) are user config
+    # mistakes: log them and exit 1, mirroring the manifest-validation handling below,
+    # rather than dumping a raw traceback through Hydra.
+    try:
+        sources = sources_from_spec({"sources": sources_dict}, key=cfg.key)
+    except (TypeError, ValueError):
```
mmcdermott: You need to forward this error to the user (and/or log it) so they know what the issue is

mmcdermott: Done in 1180850 — the handler now logs the underlying error message itself (`logger.error(f"Could not construct sources from the spec at {spec_fp}: {e}")") instead of `logger.exception`'s generic line + traceback, so the actual spec problem (e.g. `Unknown source type 's3'`) is front-and-center in the log line. Applied the same treatment to the parallel manifest-validation handler for consistency. The existing test already pins that the underlying error text surfaces (`assert "Unknown source type" in combined`) and stays green.

PR #165, opened 2026-07-29
`_metadata` as a dftly block: name-matched join keys, expression-based renaming (#146)

mmcdermott:
Implements the design in #146: the `_metadata` block is now a small **dftly program over the raw metadata table**, unifying key sourcing, normalization, metadata extraction, and `parent_codes` under the one expression language the rest of MESSY already uses.

## The contract

A `_metadata` entry maps metadata-table prefixes to a mapping of **output column name → dftly expression**, evaluated over the raw metadata table:

```yaml
chartevents:
  chart:
    code: 'f"CHART//{$itemid}"'
    _metadata:
      d_items:
        itemid: $omop_source_code   # join key: name matches a code component
        description: $label         # metadata output
        vocab: '"MIMIC-IV"'         # full dftly: literals, ??, casts, f-strings
```

- **Name matching decides the join**: produced columns whose names match the declaring event's code-referenced component columns are the **join keys**; every other produced column is metadata output attached to the matched codes.
- **Partial match** = producing a subset of the component columns (broadcasts to every code sharing the produced keys). Producing **zero** component-named columns is a config error listing the components available on that event; producing *only* keys (no outputs) errors too.
- **Renaming is just an expression** (`itemid: $omop_source_code`), and **normalization is user-visible dftly** (`::str`, `??` coalescing) — the raw-values policy from #144.
- **Reserved output names**: `code` and `code_template` may not be produced as metadata outputs. One nuance forced by the new contract: a produced column named `code` is allowed **as a join key** exactly when the code expression references a source column literally named `code` (the idiomatic ICD/OMOP vocabulary shape, #110) — keys are consumed by the component join and never collide with the pipeline-generated output `code`. `code_template` stays reserved unconditionally.
- **Validation at construction**: every block is compiled and validated by `compile_metadata_block` inside `EventConfig.__post_init__`, so config mistakes (zero keys, reserved names, unparsable expressions, leftover `_match_on`) fail at MESSY load time in every stage, as `ValueError`s with actionable messages — not mid-way through `extract_code_metadata`.

## What replaced `_match_on`

`_match_on` is **removed entirely** (0.7.0 is a breaking cut). Its validation ladder in `resolve_match_columns` is gone; match columns are now derived as the intersection of the block's produced names with the event's code-referenced components. A leftover `_match_on` key gets a targeted error pointing at MIGRATION.md, which documents the rewrite with before/after:

```yaml
# Before                                   # After
_metadata:                                 _metadata:
  med_classes:                               med_classes:
    _match_on: medication_name                 medication_name: $medication_name
    description: drug_class                    description: $drug_class
```

`extract_metadata` correspondingly loses its match-mode branching: it compiles the block through the same shared function and selects the expressions over the raw table. The reducer join path (component join, `nulls_equal`, `source_block` scoping, and #164's all-observed-codes expansion) is untouched except for the one parent_codes aggregation adjustment described below.

## Breaking: `_metadata` values have exactly dftly's semantics — a bare string is a LITERAL

Per review direction, there is **no normalization shim**: `_metadata` expressions parse exactly as dftly parses them. dftly parses a bare, unquoted word as a string *literal* (`Parser()("label")` → `Literal('label')`), so the 0.6.x shorthand `description: label` now means the constant **text** `"label"` — the raw-column read must be written `description: $label`. This is an accepted breaking change, pinned by `test_bare_string_metadata_value_is_a_dftly_string_literal` and an `extract_metadata` doctest.

The pitfall is that this is the one migration miss that **does not error**: an unmigrated bare string silently emits literal text as metadata, and a join key written as `itemid: itemid` becomes the constant `"itemid"`, which matches (at most) nothing — surfacing only as a zero-match WARNING. MIGRATION.md carries a loud dedicated warning block about exactly this.

**Legacy-form rejections (kept, and why they're legitimate):** both retained targeted errors reject shapes that are **not valid dftly anyway** (verified: `Parser()` raises "No matching node found" for each), so they are better messages on invalid input, never a reinterpretation or blocking of valid dftly:

- a **list** value (the 0.6.x coalesce shorthand / `parent_codes` matcher list) → error suggesting `coalesce($a, $b)` or a conditional, pointing at MIGRATION.md;
- the 0.6.x `{col}` interpolation string and the `{template: {matcher}}` dict fail dftly parsing and surface through the generic wrapped error naming the offending output column.

## `parent_codes` is now an ordinary dftly output (machinery retired)

The bespoke template/matcher machinery is gone from `_metadata` handling: `parent_codes` compiles through the **same** `compile_metadata_block` path as every other column (`cfg_to_expr` had no other callers in this repo and is no longer imported). The multi-case matcher shape is a chained Python-style conditional — verified against the installed dftly (`str_form/grammar.lark`: `?conditional: expr IF expr (ELSE expr)?`; `nodes/conditional.py`):

```yaml
# Before (0.6.x)                                # After (0.7.0)
parent_codes:                                   parent_codes: >-
  - "ICD{icd_version}CM/{code}":                  f"ICD{$icd_version}CM/{$code}" if $icd_version == "9"
      {icd_version: 9}                            else f"ICD{$icd_version}CM/{$code}" if $icd_version == "10"
  - "ICD{icd_version}CM/{code}":
      {icd_version: 10}
```

- **Idiomatic no-match form**: omit the final `else` — the Conditional node yields a **real null** for unmatched rows. (A bare `else null` is a trap: dftly reads the bare word as the *string* `"null"` and warns; documented in README + MIGRATION.md.)
- **Reducer compatibility**: a dftly expression yields at most one parent per metadata row, so the mapper now emits `parent_codes` as a nullable scalar String (mapper-level mandated type; `MEDS_METADATA_MANDATORY_TYPES` still documents the final `List(String)` shape). The reducer's parent_codes aggregation branch dispatches on dtype — scalar-String rows aggregate directly, a List-typed column (e.g. a shard written by an older mapper) is still flattened first — and in both cases produces the canonical per-code `List(String)`. No other reducer changes.
- **Accepted simplification**: one expression yields at most **one** parent per metadata row; a single row needing several simultaneous parents is no longer expressible in one entry. The reducer still unions parents across rows and across `_metadata` sources per code, which covers the known vocabulary shapes; a second `_metadata` entry over the same table covers the exotic case.
- New end-to-end tests: chained-ternary ICD9/ICD10 mapping, no-match-yields-null (null, never `[]`), and construction-time rejection of both legacy `parent_codes` forms.

## Deferred (out of scope, per the issue's open questions)

- **Multi-event shared metadata blocks (open question 2)**: not addressed.
- **`normalize_join_key` shrink**: the canonical-String join-side normalization (float-rendering special cases included) is kept unchanged on both sides as a safety net. Now that dftly casts give authors explicit control over key representations, this heuristic layer could shrink or disappear in a follow-up once match-coverage diagnostics (#138) make silent misses visible.

## Invariant: outputs are unchanged

Translating an existing config to the new syntax produces **identical pipeline outputs** — only the config surface moves. All committed goldens (`example/expected_output`, stage-example `out_*.yaml` fixtures) pass byte-identical with zero regeneration — including the README `parent_codes` pipeline example, whose scalar-String → reducer-aggregated `List(String)` output is value-identical to the old list-typed path; every README metadata doctest runs the real pipeline and passes.

## Test plan

- Full suite green with all extras: 186 passed (several test functions converted to doctests per review) (includes `--doctest-modules` and README/`*.md` doctests); `uv run pre-commit run --all-files` stable across consecutive runs.
- Coverage: key/output classification incl. partial match and component-named-output-is-a-key semantics; zero-join-key and `_match_on` migration errors (both at `MessyConfig.parse` and through the stage); reserved-name errors incl. the `code`-as-key allowance; bare-string-literal semantics pin; dftly key rename (`itemid: $omop_source_code`) end-to-end; `??` key normalization affecting the join end-to-end; parent_codes conditional end-to-end + null-case + legacy-form rejections; construction-time validation lives as doctests on `EventConfig.parse` / `MessyConfig.parse` / `compile_metadata_block` (per review).
- Built on post-#162/post-#164 `dev` (the all-observed-codes reducer); no overlap remains.

Refs #146


tests/test_config.py
```diff
@@ -443,3 +443,117 @@ def test_messy_save_source_deleted_errors(tmp_path):
     cfg_fp.unlink()
     with pytest.raises(FileNotFoundError, match="no longer exists"):
         cfg.save(tmp_path / "out.yaml")
+
+
+# ── _metadata blocks: validation-at-construction (dftly program over the metadata table) ──
+
+
+def test_metadata_block_validated_at_messy_parse_time():
```
mmcdermott: All of these new tests should be doctests on the associated classes.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -186,139 +196,129 @@ def _parse_code(code_value: str | NodeBase) -> tuple[NodeBase, str]:
 def resolve_match_columns(event_cfg: dict) -> list[str]:
```
mmcdermott: I'm not sure you need this function anymore with the compile base function -- it seems like this function is only used once in the code, and it is very short. Is there a better designed way to do this? Or am I wrong?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -510,51 +524,42 @@ def extract_metadata(
...
-    final_cols = []
-    needed_cols = set()
-    for out_col, in_cfg in metadata_cfg.items():
-        in_expr, needed = cfg_to_expr(in_cfg)
-        df_select_exprs[out_col] = in_expr
-        final_cols.append(out_col)
-        needed_cols.update(needed)
+    metadata_cfg = event_cfg["_metadata"]
+    if isinstance(metadata_cfg, DictConfig):
+        metadata_cfg = OmegaConf.to_container(metadata_cfg, resolve=True)
```
mmcdermott: Why do you want / need to resolve this to an OmegaConf container? Is that actually going to help us at all, or does it introduce functionality we may not want that is not currently tested?

src/MEDS_extract/config.py
```diff
@@ -46,6 +47,286 @@
...
+            f"_metadata block{ctx} uses '_match_on', which was removed in 0.7.0: join keys are "
+            f"now the produced columns whose names match the code-referenced components. Produce "
+            f"the key column(s) directly (e.g. 'itemid: $itemid') and delete '_match_on'. "
+            f"See MIGRATION.md."
+        )
+
+    component_cols = frozenset(component_cols)
+    if not component_cols:
+        raise ValueError(
+            f"The code expression {code_template_str!r} is a literal: it references no source "
```
mmcdermott, line 294: I'm realizing now that this is a limitation. What do I do if I want to indicate metadata for literal codes in my dataset? E.g., MEDS_DEATH, MEDS_BIRTH? May be worth adding an issue for tracking that in the future (though we won't address it here).

src/MEDS_extract/config.py
```diff
@@ -46,6 +47,286 @@
...
+    if not key_cols:
+        raise ValueError(
+            f"_metadata block{ctx} produces no join-key columns: none of its produced column "
+            f"names {sorted(block)} match the code expression's component columns. At least one "
+            f"produced column must be named after a component to serve as a join key. Component "
+            f"columns available on this event: {sorted(component_cols)} (from code expression "
+            f"{code_template_str!r})."
+        )
+
+    output_cols = tuple(k for k in block if k not in key_cols)
```
mmcdermott, line 345: I'm wondering if we should retain the output columns that are key_columns as well. It will help trace back provenance, though we'd need to put them in a struct like `code_components` in the raw extraction. E.g., `matched_code_components` or something. Thoughts on that?

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -32,6 +31,17 @@
     CodeMetadataSchema.parent_codes_name: pl.List(pl.String),
 }
 
+# Mapper-level mandated dtypes for the MEDS-sentinel columns a ``_metadata`` block can
+# produce. ``parent_codes``' *final* MEDS shape is ``List(String)`` (see
+# ``MEDS_METADATA_MANDATORY_TYPES``), but a dftly expression yields at most one parent
+# per metadata ROW, so the mapper emits a nullable scalar String and the reducer
+# aggregates the per-row scalars into the canonical per-code ``List(String)``.
+_MAPPER_MANDATORY_TYPES = {
```
mmcdermott: Is there a way we can just use the CodeMetadataSchema directly, rather than defining this separately? Or does it only output pyarrow types, so we can't map it to polars types?

src/MEDS_extract/config.py
mmcdermott: Agreed — filed #166 to track it. Two candidate shapes sketched there: (1) relax the zero-join-key error for pure-literal-code events so a block's columns are all outputs attached by constant to that one code, or (2) an inline top-level literal-code→metadata map (no table scan at all). Note that post-#164 the literal codes themselves DO already appear in codes.parquet with null metadata (the observed-universe seed), so the gap is purely attaching metadata values to them. Not addressed in this PR per your comment.

src/MEDS_extract/config.py
mmcdermott: I think the capability is right but this layer is the wrong place to build it. Within one block a code's matched key values are exactly its component values for the produced key columns — so round-tripping them through the metadata rows adds nothing over exporting the code-component map itself, which is already #138's scope ("export components"). Doing it on the metadata side also gets awkward at the reducer: a code can receive metadata from multiple blocks with different key sets (and partial matches produce key subsets), so a single `matched_code_components` struct has no one schema — you'd need per-block namespacing or a unified sparse struct, which the component-map export sidesteps entirely (one struct per code, full components, no join round-trip). Proposal: fold "matched_code_components in codes.parquet (or an exported components table keyed by code)" into #138 and design it together with the match-coverage diagnostics, which need the same map anyway. I've added a note to #138.

tests/test_config.py
mmcdermott: Converted in 691e7c8. All seven construction-time validation tests from this section are now doctests on the classes they document: the zero-join-key load-time failure on `MessyConfig.parse`; literal-code, `_match_on`, reserved-name, list-value, and the code-as-key allowance on `EventConfig.parse`; and the unparsable-expression / legacy `parent_codes`-form rejections on `compile_metadata_block` (which already carried the full error catalog verbatim — the `EventConfig`/`MessyConfig` doctests demonstrate the construction-time firing and context-naming, and point there for the full messages). The legacy parent_codes-form rejection test in `tests/test_extract_code_metadata.py` was converted the same way. What deliberately stays in test files: the end-to-end stage scenarios (full mapper/reducer runs over synthetic shards via `_run_ecm_scenario` — multi-file setup, worker configs, and byte-level assertions that would make unreadable doctests). `yaml_disk` wasn't needed for these: they're plain-dict `parse`-level examples, consistent with the existing `events_by_metadata_prefix`/`needed_source_columns` doctests; the README's metadata sections already use `yaml_disk` for the file-backed walkthroughs. All converted doctests run under the suite's `--doctest-modules` config (verified in the full run).

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: You're right — deleted in 691e7c8. It actually had two call sites (`resolve_match_columns` and `extract_metadata`), but each immediately followed the parse with the same `compile_metadata_block` call, so the real seam is one level up: both now go through a single `_compile_metadata_entry(event_cfg) -> (CompiledMetadataBlock, code_template_str)` helper that parses the entry's code and compiles/validates its block in one place — the two paths can no longer drift. The str-vs-NodeBase branch survives inside that seam (rather than being normalized away at the config layer) because it's inherent to the entry contract: `events_by_metadata_prefix` deliberately passes the raw code *string* when available so `code_template` stays human-readable, falling back to the parsed node only when the raw string wasn't retained — and keeping `{"code": "...", "_metadata": {...}}` as the public `extract_metadata` input keeps its doctests/direct calls ergonomic.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Good catch — deleted in 691e7c8. Reachability check: by the time `extract_metadata` runs in the pipeline, the `_metadata` block is always a plain dict — `MessyConfig.parse` converts the loaded config via `OmegaConf.to_container(resolve=True)` before `EventConfig` ever sees it, and direct/doctest callers pass plain dicts — so the branch was dead-in-pipeline defensive code, and its `resolve=True` was exactly the untested surface you suspected (interpolation resolution at stage time, with no test pinning it). Resolution semantics now live in the one place that already owns them (`MessyConfig.parse`); `compile_metadata_block` accepts any `Mapping` (which `DictConfig` is), so even an out-of-contract DictConfig input degrades to a clear parse error rather than silently resolving interpolations. A comment at the site records the invariant, and the `OmegaConf` import is gone from the stage.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: Yes — done in 691e7c8. `meds.CodeMetadataSchema` (0.4.1) exposes `.schema()` returning a pyarrow `Schema` (`code: string, description: string, parent_codes: list<string>`), and the pyarrow→polars dtype mapping is mechanical via an empty-table round-trip: `MEDS_METADATA_MANDATORY_TYPES = dict(pl.from_arrow(CodeMetadataSchema.schema().empty_table()).schema)` — which yields exactly the previous hand-written `{code: String, description: String, parent_codes: List(String)}`, so a MEDS schema change now propagates automatically and the TODO is retired. The mapper-level dict can't be fully schema-derived — `parent_codes` deliberately diverges there (a dftly expression yields at most one parent per metadata row, so the mapper emits a nullable scalar `String` and the reducer aggregates to the canonical per-code `List(String)`) — so it's now derived from the schema-derived dict with that single commented override: `_MAPPER_MANDATORY_TYPES = {**MEDS_METADATA_MANDATORY_TYPES, parent_codes_name: pl.String}`.

src/MEDS_extract/config.py
mmcdermott: I think 138 has some major issues which I'll comment on the issue, but your point about the schema being inconsistent is fair, so let's defer this for now.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
mmcdermott: I think you answered the wrong question re the function that may not be needed anymore. I wasn't talking about `parse_code`, I was talking about
  `resolve_match_columns`. It may be that you didn't need parse_code either, but my question was about the resolve_match_columns function.

mmcdermott: Apologies — my previous reply misread this thread as being about `_parse_code`; the comment was on `resolve_match_columns`. Answering the actual question: no, it was not needed, and it's gone in 13bca53. It was a thin wrapper (parse code → compile block → return `key_cols`) with a single caller in `main()`'s join-bookkeeping loop — and the smell ran deeper: the same entry was compiled twice per worker (once there for `match_cols`, once again inside the mapper's `extract_metadata`). Now `main()` compiles each entry exactly once through the `_compile_metadata_entry` seam, takes `match_cols` from `compiled.key_cols`, and hands the compiled block to the mapper via the `partial` — so per entry it's one compile at MESSY load (the `EventConfig` construction-time validation) plus one in `main()` per worker, and the mapper does zero. The compile still happens *before* `rwlock_wrap`, so the validate-in-every-worker-even-when-the-shard-exists behavior is preserved. `extract_metadata`'s signature is now `(metadata_df, compiled: CompiledMetadataBlock)` — I took the simpler-design route over an accept-either shim: its raw-dict surface (and the config-shape validation it duplicated) is gone, its doctests are rebuilt on the compiled-entry contract (`CompiledMetadataBlock` now carries `code_template`, so nothing threads the template string alongside), and the README error examples call `compile_metadata_block` directly, which is where those checks actually live.

src/MEDS_extract/extract_code_metadata/extract_code_metadata.py
```diff
@@ -170,183 +181,120 @@ def build_code_component_map(all_data: pl.LazyFrame) -> pl.DataFrame:
...
 
-    aliased = sorted(set(match_on) & {k for k in metadata_cfg if k != "_match_on"})
-    if aliased:
-        raise ValueError(
-            f"Match column(s) {aliased} may not also be declared as _metadata output "
-            "expressions. Join keys must be raw metadata columns whose names match the "
-            "code components they join against."
-        )
+    code_value = event_cfg["code"]
+    if isinstance(code_value, NodeBase):
```
mmcdermott: Do you have any doctests where this is the case? Is this ever going to be the case in practice, or is this unnecessarily defensive?

mmcdermott: It was unnecessarily defensive — gone in de52a9d. Investigation: no doctest or test exercised the NodeBase branch, and every real pipeline path takes the string side — `events_by_metadata_prefix` passes `event.raw_code`, which is set whenever the config's `code` is a string (i.e. always, for YAML configs); the goldens' human-readable `code_template` values confirm it (had the node branch ever fired, `repr(node)` — which is not valid dftly — would have leaked into them). The one reachable corner was an event writing `code` in dftly's *expanded dict form* while declaring `_metadata`: `events_by_metadata_prefix` then fell back to the parsed node. Rather than keep the branch for that corner, it's now closed upstream per the raw-string-is-canonical argument: `EventConfig` construction requires a metadata-carrying event to retain its raw code string (clear `ValueError` otherwise — `code_template` is defined as the raw dftly source, which a parsed node cannot faithfully render), so `events_by_metadata_prefix` unconditionally passes the string, and `_compile_metadata_entry` takes the raw string only (a `TypeError` spelling out the invariant if handed anything else — doctested, alongside a new `EventConfig` doctest pinning the construction-time requirement). `NodeBase` is no longer imported by the stage.

PR #185, opened 2026-08-04
meds-extract-run: forward the children's knobs instead of a fixed argv

mmcdermott:
Closes #179, closes #180, closes #181, closes #182. Refs #178.

`meds-extract-run` is a shuttle over `meds-extract-download` and `MEDS_transform-pipeline`, but it spawned each with a hard-coded argv — and offered no back door, since `RunConfig` is a structured Hydra dataclass that rejects unknown keys. This adds five passthroughs with no new semantics.

| New `RunConfig` field | Forwarded as | Issue |
| --- | --- | --- |
| `stage_runner_fp` | `MEDS_transform-pipeline --stage_runner_fp` | #179 |
| `do_profile` | `MEDS_transform-pipeline --do_profile` | #180 |
| `overrides` | `MEDS_transform-pipeline --overrides ...` | #182 |
| `download_concurrency` | `meds-extract-download concurrency=` | #181 |
| `download_continue_on_error` | `meds-extract-download continue_on_error=` | #181 |

## Notes on the shape

- **`stage_runner_fp` over an `etl:` option**, per #179's reasoning: a worker count is a property of the machine, not the dataset, and a registered spec ships inside a wheel. It is also strictly more expressive (per-stage `parallelize` and `script`).
- **`download_*` prefixed**, per #181, so the child's flat namespace doesn't leak into the runner's.
- **`stage_runner_fp` gets `user_path` absolutization** alongside the directory fields — it is user-typed and Hydra has already changed CWD by the time it is read.
- **`--overrides` must come last** (it is `nargs="*"`, so a flag after it is swallowed as another override). Pinned by doctest.

## Testing

Argv construction moved out of `main` into `download_argv` / `pipeline_argv` — two pure functions whose doctests pin the exact spellings. That alone would only prove we're consistent with ourselves, so there's also an end-to-end test running the real console script with all five knobs set: a wrong flag name makes argparse (or the download CLI's Hydra) fail the run, so a green run is the assertion.

213 tests pass; pre-commit clean.

## One usability wrinkle, documented

`overrides=` elements need quoting, since the values contain `=` which Hydra's override grammar otherwise rejects:

```bash
meds-extract-run ... "overrides=['seed=2','do_overwrite=True']"
```

README gains a passthrough table and a worked parallelism example.

### DIFF
--- a/src/MEDS_extract/run/cli.py
+++ b/src/MEDS_extract/run/cli.py
@@ -35,6 +35,7 @@ import shutil
 import subprocess
 import sys
 import sysconfig
+from dataclasses import field
 from pathlib import Path
 
 import hydra
@@ -129,6 +130,35 @@ class RunConfig:
         do_overwrite: If ``True``, the download stage re-fetches files even when
             the local copy matches.
 
+    Passthroughs to the two children. The runner is a shuttle, so these add no
+    semantics of their own — each is forwarded verbatim and documented by the child
+    that consumes it:
+
+    Fields:
+        stage_runner_fp: Path to a MEDS-transforms stage-runner file, forwarded as
+            ``--stage_runner_fp``. This is the parallelism knob: the runner reads a
+            top-level ``parallelize`` block out of that file as every stage's default
+            (and it may override ``parallelize`` or ``script`` per stage), so
+            ``parallelize: {n_workers: 8, launcher: joblib}`` in a two-line file makes
+            a whole pipeline parallel. Deliberately a *runner* argument rather than an
+            ``etl:`` option: worker counts are a property of the machine, not of the
+            dataset, and a registered spec ships inside a wheel.
+        do_profile: Forwarded as ``--do_profile``. The pipeline runner consumes this
+            itself and appends the profiler callback to each stage command, so it is
+            reachable no other way (it is neither a pipeline-config key nor a Hydra
+            override of that entry point).
+        overrides: Extra pipeline-config overrides, forwarded as ``--overrides``.
+            The escape valve for every pipeline key the synthesized config does not
+            template — ``seed``, pipeline-level ``do_overwrite``, and anything a future
+            MEDS-transforms release adds — so new keys need no change here. Quote each
+            element on the command line, since the values themselves contain ``=``::
+
+                meds-extract-run ... "overrides=['seed=2','do_overwrite=True']"
+        download_concurrency: Max parallel transport streams for the download child.
+        download_continue_on_error: If ``True``, per-file download failures don't sink
+            the run; every source is still attempted and the child exits non-zero at
+            the end if anything failed.
+
     ``__post_init__`` runs when the CLI materializes the Hydra config via
     ``OmegaConf.to_object`` (Hydra itself always hands the task function a
     ``DictConfig``; ``to_object`` is the idiomatic bridge back to the registered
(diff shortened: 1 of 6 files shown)


src/MEDS_extract/run/cli.py
```diff
@@ -178,6 +215,79 @@ def effective_input_dir(self) -> Path:
         return Path(self.download_dest_dir) if self.download_dest_dir else self.work_dir / "raw_input"
 
 
+def download_argv(run: RunConfig, spec_ref: str) -> list[str]:
```
mmcdermott: Is there a reason not have this and the other _argv method be member functions on the RunConfig dataclass? That would simplify the generated documentation for this, I think, right? It is not a big deal, though.

src/MEDS_extract/run/cli.py
```diff
@@ -178,6 +215,79 @@ def effective_input_dir(self) -> Path:
...
+    """Build the ``meds-extract-download`` child command line.
+
+    Examples:
+        >>> run = RunConfig(spec="Example", output_dir="/data/out", download_key="demo")
+        >>> for arg in download_argv(run, "pkg://ex.messy.yaml"):
```
mmcdermott: Just have this show the list like you do in the below doctest -- no need to have so many lines.

tests/test_run.py
```diff
@@ -188,3 +188,40 @@ def test_run_cli_download_free_run_and_pipeline_failure_propagates(tmp_path):
     assert result.returncode != 0
     combined = result.stdout + result.stderr
     assert "skipping the download stage" in combined
+
+
+def test_run_cli_forwards_child_flags_end_to_end(tmp_path):
```
mmcdermott, line 193: This is good, but I want you to actually test the real example with some of these knobs changed, so we know they are the right knobs to pass downstream and guard against regressions in MEDS-transforms. E.g., pass a stage runner, etc.

src/MEDS_extract/run/cli.py
mmcdermott: No good reason — done in b36b3b6. They only ever read config fields, so `self` was already all they needed, and `RunConfig` already carries `work_dir` / `effective_input_dir`, so methods match the existing shape. Call sites are now `run.download_argv(messy.spec_ref)` / `run.pipeline_argv(pipeline_fp)`.

src/MEDS_extract/run/cli.py
mmcdermott: Agreed — changed in b36b3b6. It now reads:

```
>>> run.download_argv("pkg://ex.messy.yaml")
['meds-extract-download', 'spec=pkg://ex.messy.yaml',
 'output_dir=/data/out/.meds_extract_run/raw_input', 'key=demo', 'do_overwrite=False',
 'concurrency=4', 'continue_on_error=False',
 'hydra.run.dir=/data/out/.meds_extract_run/hydra_download']
```

(The per-line loop was me avoiding a long repr; `NORMALIZE_WHITESPACE` is already on, so wrapping the list works fine — same as the `pipeline_argv` doctest.)

tests/test_run.py
mmcdermott: Done in b36b3b6 — and you were right that the test as written was too weak, in a way I only found by trying to break it.

`test_run_example.py` (the real `example/` walkthrough with golden comparison) is now parametrized over `defaults` / `passthrough-knobs`. The knobs case passes a stage runner carrying `parallelize: {n_workers: 2, launcher: joblib}` and `overrides=['do_overwrite=True']`, then runs the same golden comparison.

The catch: **matching goldens proves nothing here.** A runner that silently ignored `--stage_runner_fp` produces byte-identical output — it just runs serially. So the knobs case also asserts the flag was *honored*:

- `--multirun`, `worker="range(0,2)"` and `hydra/launcher=joblib` appear in the pipeline log (i.e. the pipeline runner built the parallel stage command);
- `shard_events/.logs/` contains two per-worker dirs `0` and `1` (i.e. two workers actually ran).

Verified by mutation — replacing the `--stage_runner_fp` append in `pipeline_argv` with a no-op leaves the run green *and* the goldens matching, and fails only on:

```
AssertionError: stage_runner_fp not honored: '--multirun' absent
```

which is the MEDS-transforms regression you were asking to guard against.

One deliberate omission: `do_profile` stays out of the example run, since it needs `hydra_profiler` installed — exercising it there would test the plugin's availability rather than our passthrough. It stays covered at the argv level.

PR #260, opened 2026-08-05
merge_to_MEDS_cohort: multithreaded stable sort, unique_by=null default, property-tested

mmcdermott:
Implements #241 with the maintainer-requested property-test safeguard.

**Changes:**
1. The per-shard sort is now `multithreaded=True`, keeping `maintain_order=True` — a stable sort's output is input-determined regardless of thread count, so determinism is unaffected (and was verified byte-identical).
2. The shipped default `unique_by` flips `"*"` → `null`. The dedup was provably a no-op in the shipped pipeline: `EventConfig.extract` ends every event block with a full-width `.unique(maintain_order=True)` and stamps a per-block `source_block`, so rows within a block are pre-deduped and rows from different blocks always differ. `"*"` remains available as opt-in belt-and-braces; docstring and config comment document the invariant.

**Property tests** (`tests/test_merge_properties.py`, hypothesis): generators build pipeline-shaped inputs (per-prefix frames from 1–3 blocks, tie-heavy time pools including nulls, optional per-prefix `code_components` structs, per-block dedup mirroring upstream), always chained through the real `_filter_to_subjects` then the real `merge_subdirs_and_sort`. Asserted per example: **EQUIVALENCE** — new-settings output equals a pre-#241 oracle (old concat → `unique("*")` → single-threaded stable sort) exactly, `check_row_order=True`, and the `unique_by="*"` opt-in path stays byte-identical too; **SORTED** (stable re-sort is identity); **UNIQUE**; **LOSSLESS** (row count = sum of inputs). 50 examples in CI (2.8s), 300 via `HYPOTHESIS_PROFILE=thorough`; plus deterministic tie-heavy / empty-prefix / single-subject regressions.

**Measured** (4M rows, 4 tables, one shard): 3.6–3.9s → 1.1–1.4s (~2.6–3.6×); attribution ~0.5s from dropping the no-op dedup, ~1.5–2s from sort threading. Outputs identical including row order.

Also adds `hypothesis` to the dev group (note: #255 adds the same line — whichever merges second needs a trivial `uv lock` regen).

Closes #241.


src/MEDS_extract/merge_to_MEDS_cohort/config.yaml
```diff
@@ -1,2 +1,9 @@
-unique_by: "*"
+# `null` skips post-merge deduplication entirely. That is safe for pipeline-produced inputs
```
mmcdermott: I think this comment probably won't surface in user-facing documentation as naturally as something in a README for this stage or in a docstring/doctest, so it probably should be moved.

src/MEDS_extract/merge_to_MEDS_cohort/merge_to_MEDS_cohort.py
```diff
@@ -290,7 +290,10 @@ def merge_subdirs_and_sort(
             else:
                 logger.warning(f"Column {s} not found in dataframe. Omitting from sort-by list.")
 
-    return df.sort(by=sort_by, maintain_order=True, multithreaded=False)
+    # `maintain_order=True` makes this a stable sort, and a stable sort's output is fully determined
+    # by its input regardless of how many threads execute it — so multithreading changes nothing
+    # about the result while being ~2x faster on large shards (#241).
```
mmcdermott: don't reference issues that will be stale or design decisions that people wouldn't question without historical context that will be outdated -- so this whole comment can go.

tests/test_merge_properties.py
```diff
@@ -0,0 +1,218 @@
...
+            _block_frame(prefix, block_idx, draw(st.lists(row, max_size=12)), with_components)
+            for block_idx in range(n_blocks)
+        ]
+        frames[prefix] = pl.concat(blocks, how="vertical")
+
+    subjects = sorted(draw(st.sets(st.sampled_from(SUBJECT_POOL), min_size=1)))
+    return prefixes, frames, subjects
+
+
+def _pre_241_reference(frames: list[pl.DataFrame]) -> pl.DataFrame:
```
mmcdermott: Call this something else -- we don't want to reference an issue that will become a stale reference over time. This is a guaranteed deterministically ordered output, that's all.

tests/test_merge_properties.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Diagonal-relaxed concat in prefix order, full-column ``unique(maintain_order=True)``
+    (the old ``unique_by: "*"`` default), then the old single-threaded stable sort.
+    """
+    lf = pl.concat([f.lazy() for f in frames], how="diagonal_relaxed")
+    lf = lf.unique(maintain_order=True)
+    return lf.sort(by=["subject_id", "time"], maintain_order=True, multithreaded=False).collect()
+
+
+def _check_merge_properties(prefixes: list[str], frames: dict[str, pl.DataFrame]) -> None:
```
mmcdermott, line 136: If this is possible in a meaningful way, it may also be worth testing stability -- e.g., given identical inputs, if you do this twice is the output consistent? But if the concern on randomness is somehow python-session or time based or something, that may not be viable, in which case you can ignore.

tests/test_merge_properties.py
```diff
@@ -0,0 +1,218 @@
...
+
+settings.register_profile("default", max_examples=50, deadline=None)
+settings.register_profile("thorough", max_examples=300, deadline=None)
+settings.load_profile(os.environ.get("HYPOTHESIS_PROFILE", "default"))
+
+SUBJECT_POOL = [1, 2, 3, 4, 5]
+
+# Small pools (plus None for static events) so ties on (subject_id, time) are common — ties across
+# blocks are exactly where instability in the merge/sort would surface.
+TIME_POOL = [
```
mmcdermott: you want these to be true continuous distributions. The goal of a property test is coverage / breadth, not exhaustive checking. These types are easy, so this can be easily done.

src/MEDS_extract/merge_to_MEDS_cohort/config.yaml
mmcdermott: Moved in 599cc88: config.yaml now carries a one-line pointer, and the full rationale lives in the `main` stage docstring's `unique_by` description (which feeds the generated stage docs), where it was already partially present.

src/MEDS_extract/merge_to_MEDS_cohort/merge_to_MEDS_cohort.py
mmcdermott: Deleted in 599cc88 — and the same rule (no issue references or historical-context design notes in code comments) was applied across the whole diff: the test module docstring and oracle docstring are also scrubbed. I'll sweep the stacked PRs (#262/#265) the same way when syncing them.

tests/test_merge_properties.py
mmcdermott: Renamed in 599cc88 to `_deterministic_merge_reference`, documented as what it is: the guaranteed deterministically ordered merge (diagonal concat in config order, full-column dedup, single-threaded stable sort) that any settings must reproduce exactly.

tests/test_merge_properties.py
mmcdermott: Viable and added in 599cc88 as the STABLE property: every generated example now merges the same on-disk inputs twice and asserts the frames are identical including row order (check_exact). There is no session/time dependence in the plan — stable sort + maintain_order unique are input-determined — so this is a real assertion, and it would catch any future nondeterminism regression (e.g. an engine change) immediately.

tests/test_merge_properties.py
mmcdermott: Redesigned in 599cc88, same philosophy as the #255 rework: every value now draws from its full type domain — subject ids and struct ints from all of Int64, times via st.datetimes() (plus nulls), codes/text from arbitrary unicode, numerics from all of Float32 including NaN/inf, and prefixes/struct field names are drawn identifiers with per-prefix struct shapes and dtypes varying. The one structural constraint kept: rows sample from a small per-example pool of those full-domain draws, because (subject_id, time) ties across blocks are where ordering bugs hide, and fully independent draws would almost never tie. The pool is redrawn from the full domain every example, so nothing is enumerable. Verified at 300 examples (thorough profile) locally.

PR #262, opened 2026-08-05
convert_to_subject_sharded: sink the shard plan instead of collecting it

mmcdermott:
Implements #240, stacked on #260 (shares its property-test suite). Gated on the requested `sink_parquet` stability investigation, whose verdict:

- **The historical errors trace to polars' OLD streaming engine** (the era when `write_lazyframe` was born eager, May 2024) — deprecated ~1.23 and deleted upstream in 1.31. Full-history search of MEDS-transforms (1,332 commits, pickaxe included) and this repo found **no recorded sink failure ever**; sinks were stabilized upstream in 1.30, and this repo already sinks in production in the ingest stage.
- Empirically (polars 1.38.1 / 1.39.3 / 1.43.2, incl. struct/categorical/null-key/zero-row/empty-shard cases): all variants content-equal everywhere; the one real hazard is **row order** — naive sink reorders nondeterministically (every run different), and that order provably leaks into final MEDS bytes through merge's stable sort for same-time events.
- With `join(maintain_order="left_right")` + `sink_parquet(maintain_order=True)`, sink output is **byte-deterministic across runs/threads/versions and equals the eager path's exact row order**, at **5.2× lower peak memory** (4250 MB → 823 MB on a 30M-row table) and fastest wall time. Works at the 1.38 floor.

**The change:** `JoinConfig.apply` joins with `maintain_order="left_right"` (a no-op for the in-memory engine — every other consumer's behavior is unchanged, verified byte-identical), and the stage's write_fn becomes a local `sink_df` with `write_df`'s exact `.tmp` + `os.replace` atomicity. This also answers the memory-semantics question from review: the plan always only *kept* shard-sized output — the eager engine just materialized the full join transiently before filtering; streaming execution makes the transient match the logical requirement.

**Tests:** the shared property suite gains the write-path contract — sunk ≡ eager including row order on generated stage plans (`check_row_order=True`), repeated sinks byte-identical — plus deterministic fan-out-join (duplicate right keys, the case an unordered streaming join is free to reorder) and empty-shard (valid zero-row schema-correct file) regressions. 59 tests green locally incl. stage examples and the in-process pipeline; pre-commit clean.

**Documented residual:** float `sum`/`mean` aggregated-join columns are ULP-nondeterministic under any streaming execution ({min,max,count} exact) — noted in the commit; can be gated later if a real config hits it.

Closes #240.


tests/test_merge_properties.py
```diff
@@ -266,3 +276,88 @@ def test_merge_properties_single_subject_shard(tmp_path: Path) -> None:
...
+    eager = lf.collect()
+    fp_a, fp_b = Path(tmpdir) / "a.parquet", Path(tmpdir) / "b.parquet"
+    sink_df(lf, fp_a)
+    sink_df(lf, fp_b)
+    assert_frame_equal(pl.read_parquet(fp_a), eager, check_row_order=True, check_exact=True)
+    assert fp_a.read_bytes() == fp_b.read_bytes(), "Repeated sinks of one plan must be byte-identical."
+
+
+@given(inputs=shard_inputs())
+def test_subject_sharded_sink_write_properties(
```
mmcdermott: I think you can merge this with the other property test, right? Then you can just test both properties at once, to not need to draw separate samples for each? Or is that an antipattern for this kind of test?

tests/test_merge_properties.py
mmcdermott: Why not let this be min_size=0 to include the empty filter setting? And/or let this also include subjects not in the subject_pool for the subject-outside-of-file case?

tests/test_merge_properties.py
mmcdermott: Merged in the new head commit — and agreed it's not an antipattern here: both families consume the same input domain along one pipeline chain, so the combined test filters each prefix, runs the write-path checks on the lazy plan, collects, and hands the frames to the merge checks. One draw exercises both; failures still localize via the assertion messages, and hypothesis shrinks the shared input normally.

tests/test_merge_properties.py
mmcdermott: Done in the new head commit — both variants: the subject subset now draws with min_size=0 (empty-shard state) and unions in up to two full-domain Int64 ids outside the pool (subject-not-in-file state), so both edge states arise organically in the search rather than only via the pinned regressions. Verified at 300 examples.

PR #265, opened 2026-08-05
Drop code_components at merge; reorder extract_code_metadata before merge (#254)

mmcdermott:
The #254 memory fix, per maintainer decision, stacked on #262. **Read the first section before anything else — it corrects the premise this fix was approved under.**

## Critical: the shipped pipeline order contradicted the documented design

`docs/ingestion_pathways.md` has always said `extract_code_metadata` consumes the `convert_to_MEDS_events` output — but the shipped pipeline configs ordered it **after** `merge_to_MEDS_cohort`, so its `data_input_dir` actually resolved to the merged shards. Dropping `code_components` at merge under that order silently degrades `metadata/codes.parquet` to a codes-only table (caught during golden regeneration). A preparatory commit therefore reorders the canonical pipeline — `extract_code_metadata` now runs between `convert_to_MEDS_events` and `merge` in `EtlConfig.DEFAULT_PIPELINE`, `configs/_extract.yaml`, `example/pipeline.yaml`, and the consistency test. As a metadata stage it doesn't advance the data chain, so merge's input is unchanged, and the produced `codes.parquet`/`subject_splits.parquet` are **byte-identical** to the checked-in goldens. **Release note needed:** any user-authored pipeline YAML keeping the old order will silently lose metadata enrichment after this change (warning logged).

## The fix

`merge_subdirs_and_sort` drops `code_components` per scan (`.drop(..., strict=False)`), so the ~70-field `diagonal_relaxed` superstruct never forms and projection pushdown never reads the column. Per maintainer ruling, rows that look identical in the merged output are **deduplicated**: the shipped `unique_by` default is restored to `"*"` with its rationale rewritten — post-drop it is a semantic guarantee (rows that differed only in components collapse; identical-looking output rows must not be duplicated), no longer a provable no-op. `null` remains the raw-concat escape hatch; #260's multithreaded stable sort stands.

**Measured, eICU-shaped synthetic (30 tables, 5.44M rows, 60 MB), shipping configuration (drop + `"*"` unique + multithreaded sort):** peak RSS **12,419 → 2,329 MB (5.3×)**, wall **41.4 → 4.9 s (~8×)**, on the repo-locked polars 1.39.3. This also resolves the #260 memory interaction (the superstruct is gone before the sort's gather). The dedup collapsed exactly the 2 synthetic rows that differed only in components.

## Alternative preserved for the record (maintainer request)

If per-row components are ever wanted in final data: **explode the struct into name-prefixed flat columns pre-concat, re-collapse post-sort** — avoids the superstruct densification while keeping the data (flat columns of mostly-null Strings are cheap; the 70-field cost came from *struct child* buffers under dense materialization), and should be comparably fast. Also measured: JSON-encoding the struct pre-concat (3.1 GB, ~6× faster, exact round-trip). Both were set aside because the column is internal metadata linkage that nothing downstream needs today (#138's export story goes through metadata).

## Tests

Property suite updated: DROPPED (components absent from output whenever inputs carry them), EQUIVALENCE against the new oracle (concat → drop → full-column unique → stable sort), SORTED, UNIQUE (unconditional full-row contract), LOSSLESS-minus-collapse; deterministic regression `test_merge_dedups_rows_that_differed_only_in_components` (same-block pair differing only in components: 2 in → 1 out; cross-block twin kept via `source_block`; `unique_by=null` keeps all 3). Final-data goldens regenerated via the real pipeline (columns now `[subject_id, time, code, numeric_value, source_block]`); merge/finalize stage examples and README/ingestion-pathways prose updated; README doctests now read components from the pre-merge tree. Full battery green: default suite, golden byte-compare, parallelism lane, README doctests; pre-commit clean.

Closes #254.


example/pipeline.yaml
```diff
@@ -22,7 +22,9 @@ stages:
   - convert_to_subject_sharded
   - convert_to_MEDS_events:
       do_dedup_text_and_numeric: false
-  - merge_to_MEDS_cohort
+  # extract_code_metadata runs BEFORE the merge: it joins metadata against the
```
mmcdermott: This comment is referencing a historical issue and not necessary.

src/MEDS_extract/configs/_extract.yaml
```diff
@@ -22,7 +22,11 @@ stages:
   - split_and_shard_subjects
   - convert_to_subject_sharded
   - convert_to_MEDS_events
-  - merge_to_MEDS_cohort
+  # extract_code_metadata runs BEFORE the merge: it joins metadata against the
```
mmcdermott: This comment is not necessary here, it references a design decision people wouldn't otherwise question and will just drift historically.

src/MEDS_extract/merge_to_MEDS_cohort/examples/default/README.md
```diff
@@ -1,4 +1,7 @@
 Concatenates the per-table event parquets (`convert_to_MEDS_events` output) within each
 `(split, shard)` into a single file at `data/<split>/<shard>.parquet`. Rows are sorted
-so null `time` values (static events like `EYE_COLOR`) precede real timestamps per subject,
-and the resulting schema unions all `code_components` struct fields across tables.
+so null `time` values (static events like `EYE_COLOR`) precede real timestamps per subject.
+The internal `code_components` struct is dropped during the merge (#254): unifying the
```
mmcdermott: don't reference issues that will be stale, or design decisions people won't question.

src/MEDS_extract/merge_to_MEDS_cohort/config.yaml
```diff
@@ -1,3 +1,3 @@
-# See the stage docstring for `unique_by` semantics and why the shipped default is safe.
-unique_by: null
+# See the stage docstring for `unique_by` semantics and why the shipped default is `"*"`.
+unique_by: "*"
```
mmcdermott, line 2: I'm not sure why this is changing back here. It seems like this could cause other problems if people are adding other columns into this. Maybe this should by default be all the required MEDS columns only? I'm not sure. Does this even need to exist, or is this needlessly defensive?

src/MEDS_extract/config.py
```diff
@@ -1791,8 +1791,12 @@ class EtlConfig:
         "split_and_shard_subjects",
         "convert_to_subject_sharded",
         "convert_to_MEDS_events",
-        "merge_to_MEDS_cohort",
+        # ``extract_code_metadata`` must run off the *pre-merge* per-table events: it joins
```
mmcdermott: ditto prior comment -- also we need to ensure this doesn't result in the loss of metadata.

tests/test_merge_properties.py
```diff
@@ -132,14 +140,16 @@ def shard_inputs(draw: st.DrawFn) -> tuple[list[str], dict[str, pl.DataFrame], l
     return prefixes, frames, subjects
 
 
-def _deterministic_merge_reference(frames: list[pl.DataFrame]) -> pl.DataFrame:
+def _deterministic_merge_reference(frames: list[pl.DataFrame], *, unique: bool) -> pl.DataFrame:
```
mmcdermott, line 146: For a true deterministic merge reference, why not collect the frames from the start and just do everything in memory? Then you'll also ensure that we test that lazy doesn't cause issues too, right? Or am I misunderstanding? Or do you want to do that separately so we test both the collect pre write vs. collect at start?

example/pipeline.yaml
mmcdermott: Deleted in the new head.

src/MEDS_extract/configs/_extract.yaml
mmcdermott: Deleted in the new head; the ordering itself stays pinned by the end-to-end golden byte-compare (a reordered pipeline produces a codes-only codes.parquet and fails it) and the DEFAULT_PIPELINE/_extract.yaml consistency test.

src/MEDS_extract/merge_to_MEDS_cohort/examples/default/README.md
mmcdermott: Cleaned in the new head: the README keeps only the user-facing fact (the merged output does not carry code_components — readers see the column present in in.yaml and absent in out_data.yaml), with the issue ref and rationale removed.

src/MEDS_extract/merge_to_MEDS_cohort/config.yaml
mmcdermott: Left as "*" deliberately — reasoning, since this one's a real design question: (1) It needs to exist: with components dropped at merge, rows that differed only in their raw components become literally identical rows, and per your earlier ruling identical-looking rows must not be duplicated — this dedup is the mechanism implementing that, no longer belt-and-braces. (2) "*" is actually the *safe* shape for user-added columns, not the risky one: dedup over all-columns-present only ever collapses rows that are fully identical, so no information is ever lost. A required-MEDS-columns-only default would be the dangerous version — it would collapse rows that differ in a user-added column and drop one of the differing values nondeterministically (the docstring already warns subset-dedup retention is unspecified). Put differently: "*" is exactly the 'looks identical in the output' criterion; a fixed subset would violate it in both directions the moment extra columns exist. Happy to revisit if you want different semantics, but I believe "*" is the faithful encoding of the rule you stated.

src/MEDS_extract/config.py
mmcdermott: Comment deleted in the new head. On metadata loss: it cannot regress silently — the end-to-end golden tests byte-compare the *enriched* metadata/codes.parquet (descriptions + parent_codes populated), and running with the wrong order produces a codes-only table that fails that compare immediately (that is exactly how this issue was caught during golden regeneration). The metadata goldens were verified byte-identical under the reorder, and extract_code_metadata's own warning still fires if anyone hand-orders it after merge.

tests/test_merge_properties.py
mmcdermott: You're reading it right, and done in the new head: the oracle now computes every operation eagerly (DataFrame concat -> unique -> single-threaded sort, no lazy plan anywhere), so the equivalence assertions differentially guard the real merge's lazy execution path against a plain in-memory computation of the same semantics — lazy-engine bugs now fail the property rather than cancel out. The collect-before-write vs collect-at-start distinction is separately covered on the write side: _check_write_path_properties compares the sunk output against an eager collect of the same plan, byte-deterministically.

README.md
```diff
@@ -301,23 +305,30 @@ shape: (7, 2)
 > yields one event, not two.
 
 The `code_components` struct column preserves the individual column values that were
-combined to form the code. This enables queries on code components without parsing the
-code string — for example, finding all Glucose readings regardless of units:
+combined to form each code, enabling queries on code components without parsing the
+code string. It lives on the *pre-merge* per-table event files — the
+`convert_to_MEDS_events` stage output, cached under the output directory — where
+metadata linking consumes it; `merge_to_MEDS_cohort` drops it from the final shards
+(#254: unifying every table's struct into one field-union superstruct is
```
mmcdermott: Drop references here and throughout to issues by number, as they'll go stale and just contribute comment/doc bloat. Especially in a README, code components don't warrant too much mention as they are now just an implementation detail, not a forward facing interface element (except for in metadata linking).

README.md
```diff
@@ -1122,7 +1137,13 @@ codes from the event block that declared it (that is what `source_block` is for)
 ...       time:
 ... ''', Path(tempfile.mkdtemp()))
 >>> run_extraction(root)
->>> data = pl.read_parquet(f"{root}/output/data/**/*.parquet")
+>>> data = pl.concat(
+...     [
+...         pl.scan_parquet(f"{root}/output/convert_to_MEDS_events/**/{table}.parquet")
```
mmcdermott, line 1139: Why doesn't the **/* still work?

README.md
mmcdermott: Done in the new head: every issue-number reference in the README is gone (grep-verified zero), the embedded pipeline snippet matches the cleaned example file, and code_components mentions outside the metadata-linking section are trimmed to the user-facing facts — the linking section keeps its full treatment as the one place it is interface.

README.md
mmcdermott: It genuinely can't: a polars multi-file scan/read requires a single unified schema across the globbed files, and the per-table code_components structs have different fields — verified empirically on our locked polars: pl.read_parquet('**/*.parquet') and scan_parquet both raise SchemaError ('data type mismatch for column code_components: ... encountered extra struct field'). Multi-file scans do not diagonal_relaxed-unify. The old snippet's single glob only worked because it read the *merged* output, where the superstruct had already been unified (the very thing that made merge memory-hungry). Added a brief comment in the snippet so the next reader doesn't wonder the same thing.

